PR #46, opened 2017-08-06
accounting start/stop delay fields

pavelsly:


nemesisdesign:
Hi @pavelsly, thank you for contributing!

We haven't found these two fields in the freeradius 3 schema, [it looks like they have been removed a few years ago](https://github.com/redBorder/freeradius/commit/430a5f7ea8b70d42897427326537c6bfe1e5643d), I believe they are from freeradius 2.
Could you expand a bit why you need them? Are you testing django-freeradius on your own use case and noticed something is not working out or what?

peppelinux:
Hi @pavelsly what about the freeradius migration from 2.9 to 3.x?
We still have this pending pull request that we would like to close with your help :)
let us know

PR #61, opened 2017-12-11
RadiusCheck password auto-hash and validation, custom admin filters

peppelinux:
In this PR we have:

- RadiusCheckAdmin entry update with or without new_value field value
- RadiusCheckAdmin custom filters: find duplicates by username or value, find is_active true or false and find valid_until < or > then now()
- RadiusCheck custom model manager
- _encode_secret standalone funcion (in base/models.py), this still need a better configuration. It should not be forget into models.py but moved in an apposite file (should it be called helpers.py?)
- enhanced app_settings.py approach, following nemesys's indications
- freeradius documentation on how to do to extend radiucheck query (mysql dialect example)

Hope it will enjoy you

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
 django>=1.11,<1.12
 django-model-utils
 swapper
-djangorestframework>=3.3,<3.7
+djangorestframework>=3.7,<3.8


peppelinux:
Code coverage test done with 99% because still have a not covered code in
base/admin.py, line 29 (ReadOnlyAdmin.get_actions method). This is not my code so I think that I should wait for your personal revision on it.

Travis-ci also get a strange error in requirements-test, it have a conflicting dependencies on django version... take a look and let me know, I'm on it :)

django_freeradius/base/forms.py
```diff
@@ -0,0 +1,40 @@
...
+                                min_length=8, max_length=16,
+                                widget=forms.PasswordInput(),
+                                help_text=_secret_help_text)
+
+    def clean_attribute(self):
+        if self.data['attribute'] not in app_settings.DISABLED_SECRET_FORMATS:
+            return self.cleaned_data["attribute"]
+        # this is not covered anymore
+        # raise ValidationError('This SECRET FORMAT is disabled'
+        #                      ', please use another one')
```
nemesisdesign: can you remove this commented code please?

docs/source/general/freeradius.rst
```diff
@@ -50,27 +64,49 @@ Example for configuration with postgresql::
     password = "<password>"
     radius_db = "radius"
 
-Enable the ``sql`` and ``rest`` modules:
+Enable
```
nemesisdesign: this doesn't look right here

django_freeradius/base/forms.py
peppelinux: done

docs/source/general/freeradius.rst
peppelinux: done, stupid code and paste!

Issue #63, opened 2017-12-12
[Feature] Disabled Users in API


cappe87:
We can add a possibilty to manage disabled users.
A standard way to have disabled users is to use a group in radgroupcheck table.
Here you can see an [example](https://pastebin.com/bRtn8Rmr
)
We can decide to have a simple checkbox in the users page like "Disabled" or we can improve a simple method to add users to one or more groups. 
I prefer the last one.

peppelinux:
Hi Cappe, I developed an is_active and a valid_until attribute in RadiusCheck.
You can find my code in this PR:
https://github.com/openwisp/django-freeradius/pull/61

We could code the same behaviour for the groups.

ps = I also update the Documentation to let the user modify the behaviour of sql freeradius radcheck query. If valid_until threshold is exceeded or is_valid is false the query will ignore disabled accounts.

cappe87:
In my production freeradius I use the same "solution".
We can decide to use this adding is_active and valid_until to RadiusCheck table or try to use a standard solution with radgroupcheck. What do you think about this?

peppelinux:
is_active and valid_until is going to be merged soon, I think (I hope), in radiucheck schema.

I think that is_active and valid_until could be also used in radgroupcheck.

Doing this we could also extends the RadGroupUserAdmin with username autocompletion (I have some code roughly ready to be merged, with third-part dependency called django-autocomplete-light) because Alan Dekok didn't used ForeignKeys but string value comparisons... So we have to deal with this and store the string value of a radcheck entry instead of its ID (Can the rdbms forgive us for this?!)

peppelinux:
Merge done, master branches aligned
we could start creating a new task for CGI if you agree, the focus should be how to disable or expire a group of radcheck entries using radgroupcheck

nemesisdesign:
BTW: the best way for the classical users of OpenWISP is to do this is via the RESTful API since they don't have to mess with radius checks and can manage users directly from the users management page.
This will become especially important when multitenancy is added to the mix.

Adding this to the RESTful API is just a matter of few lines of code and we can turn this into a GCI task or let @AlessiaVe handle it if she wants.

nemesisdesign:
There: https://github.com/openwisp/django-freeradius/blob/master/django_freeradius/api/views.py#L21

Possible path forward (TDD style):

- write a test that creates a disabled user and tries authenticating it, expecting the authentication to fail, the test will fail because it doesn't seem implemented this way
[...]

Issue #35, opened 2017-07-26
[tests] Make all the tests reusable  


lillopaco:

peppelinux:
Would we like to create a unit test method to take some arguments to introduce an generalized approach?

We could introduce a method as an interface that call some other private methods of the class.
Do we start with a GCI task?

nemesisdesign:
@peppelinux we do this in other openwisp2 modules, take a look:

- https://github.com/netjson/django-netjsongraph/tree/master/django_netjsongraph/tests/base
- https://github.com/openwisp/django-loci/tree/master/django_loci/tests/base

Yes let's create it

nemesisdesign:
Closed by 79a451c402fde45afdf5277f171ca33665c35830

See @R9295, we forgot to close this ticket because the commit message didn't close it automatically.
It's important to automate this stuff because it's such a waste of time to do it manually.

R9295:
@nemesisdesign gotcha, will do next time.

PR #64, opened 2017-12-13
[RadiusCheck] Password encryption #58

### LINKED ISSUE #58: Password Encryption
Hi guys,

I'm wondering has anyone given thought to, or is it within the scope of the projects goals, to add the respective encryption algorithms for the radcheck passwords and an option to encrypt from the admin panel?

I'm going to have to do this myself anyway, for at least one algorithm, so would be happy to contribute if this is something that is welcomed.

**Edit:** To clarify, I mean to add an extra field + button to populate the **value** field with the encrypted password, as seen below:

image

peppelinux:
[RadiusCheck] Password encryption #58 Disable Users and more generalized tests

- RadiusCheckAdmin entry update with or without new_value field value
- RadiusCheckAdmin custom filters: find duplicates by username or value, find is_active true or false and find valid_until < or > then now()
- RadiusCheck custom model manager
- _encode_secret standalone funcion (in base/models.py), this still need a better configuration. It should not be forget into models.py but moved in an apposite file (should it be called helpers.py?)
- enhanced app_settings.py approach, following nemesys's indications
- freeradius documentation about how to do to extend radiucheck query (mysql dialect example)
- more generalized tests.py 

Fixes #58
Improvements for #35 and #63


peppelinux:
Ping !

should we change something? 
Is there some minor change that you would like to talk about, is this ready to be merged or you are waiting for something more to be fixed by me?
Is there something wrong with the coding style, are these changes not compliant to the project or simply we need more time to think on it? :)

PR #72, opened 2017-12-21
Fix link to project page in README.rst

okraits:
Link in README.rst goes to OpenWRT and not to OpenWISP.


h0n3yc4k3:
It already sent :) https://github.com/openwisp/django-freeradius/pull/71

Issue #59, opened 2017-12-02
[Feature] List of nas type in the nas add page


cappe87:
In the nas add page we can add a list of nas type on the right of the input box.
A free input box will remain, user can decide to write a nas type or choose one from the list.
I think it would be help.
What do you think @nemesisdesign ?

peppelinux:
These could be extracted by service-type from rfc 2865 and putted in tha app_settings as done for the hash type

https://www.iana.org/assignments/radius-types/radius-types.xhtml#radius-types-13

cappe87:
This should be fixed by https://github.com/AlessiaVe/django-freeradius/commit/12e0539fbba5e61e3c485b2fa850396fe63f152f
![add nas django site admin](https://user-images.githubusercontent.com/6532945/33935199-609fb92c-dffb-11e7-8a60-989540616849.png)

peppelinux:
Wonderfull, she just have to test flake/isort/coverage and then do a pull request !!!

PR #81, opened 2017-12-28
Add other Nas Type

AlessiaVe:
For have two fields in the admin's interface, which aren't in the model, I added a file Forms.py. In this file there is a class NasModelForm with two fields: standard_type and other_NAS_type. For saving the correct value of the type in the model I overrode the method save_model of the class AbstractNasAdmin in admin.py. I made these changes as solution of the issue #59. I did flake8, isort and coverage test.


peppelinux:
isort fails, I know your feel :)
Once the changments will be done remember to
````
git commet --amend 
````
to join and overwrite the latest two commits into one.

peppelinux:
Mitica. l'hai fatto :)

AlessiaVe:
The documentation? What file I have to change?

peppelinux:
docs/source/
here you'll find all the doc sources.
there's also a shpinx builder for validation (so: flask+isort+cover+sphinx_build)

Issue #47, opened 2017-08-08
[accounting] Dealing with unterminated radacct entries


nemesisdesign:
@yakky posed a couple of good questions, a third was then posed to myself:

1. ~~what happens if an accounting with `'Acct-Status-Type'= 'Start'` is sent but an accounting with same `Acct-Unique-Session-Id` already exists?~~ has been addressed in #45 and will be optimized in #49 
2. ~~what happens if an accounting with `'Acct-Status-Type'= 'Interim-Update'` is sent but the accounting does not exist?~~ has been addressed in #45 and will be optimized in #49 
3. (added by myself) what happens if an accounting is never stopped? It happened to us at @cineca a few times in the past and my colleage @dariomas told me they solved it with a script

Some of these points may become separate issues in the near future. Acting now without full understanding and a real world instance may be premature, but is better to start thinking about if it's worth to try to solve these problems within the application itself and if yes how and if these automation can be turned off in case of need.

nemesisdesign:
Closed by #82 
@R9295 same here

PR #82, opened 2017-12-29
[accounting] Added command to deal with unterminated radacct entries #47

R9295:
Implements n. 3 of #47

### DIFF
--- a/django_freeradius/management/commands/cleanup_stale_radacct.py
+++ b/django_freeradius/management/commands/cleanup_stale_radacct.py
@@ -1,0 +1,24 @@
+from datetime import timedelta
+
+from django.core.management import BaseCommand
+from django.utils.timezone import now
+
+from django_freeradius.models import RadiusAccounting
+
+
+class Command(BaseCommand):
+    help = "Closes active accounting sessions older than <days>"
+
+    def add_arguments(self, parser):
+        parser.add_argument('number_of_days', type=int, nargs='?', default=15)
+
+    def handle(self, *args, **options):
+        days = now() - timedelta(days=options['number_of_days'])
+        sessions = RadiusAccounting.objects.filter(start_time__lt=days, stop_time=None)
+        for session in sessions:
+            # calculate seconds in between two dates
+            session.session_time = (now() - session.start_time).total_seconds()
+            session.stop_time = now()
+            session.update_time = session.stop_time
+            session.save()
+        self.stdout.write('Closed active sessions older than {} days'.format(options['number_of_days']))
--- a/django_freeradius/tests/test_admin.py
+++ b/django_freeradius/tests/test_admin.py
@@ -17,6 +17,10 @@ _RADCHECK_ENTRY = {'username': 'Monica', 'value': 'Cam0_liX',
                    'attribute': 'NT-Password'}
 _RADCHECK_ENTRY_PW_UPDATE = {'username': 'Monica', 'new_value': 'Cam0_liX',
                              'attribute': 'NT-Password'}
+_RADACCT = {'username': 'bob', 'nas_ip_address': '127.0.0.1',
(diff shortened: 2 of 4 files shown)


django_freeradius/management/commands/cleanup_stale_radacct.py
```diff
@@ -0,0 +1,25 @@
...
+from django.utils.timezone import now
+
+from django_freeradius.models import RadiusAccounting
+
+
+class Command(BaseCommand):
+    help = "Closes active accounting sessions older than <days>"
+
+    def add_arguments(self, parser):
+        parser.add_argument('number_of_days', type=int)
```
nemesisdesign: could you add a default value of 15 days?

django_freeradius/tests/test_admin.py
```diff
@@ -213,3 +213,15 @@ def test_nas_admin_save_model(self):
         self.assertNotContains(response, 'error')
         nas.refresh_from_db()
         self.assertEqual(nas.type, 'my-custom-type')
+
+    def test_cleanup_stale_radacct_command(self):
+        RadiusAccounting.objects.create(
+            unique_id='117', username='bob', nas_ip_address='127.0.0.1', start_time='2017-06-10 10:50:00',
+            authentication='RADIUS', connection_info_start='f', connection_info_stop='hgh',
+            input_octets='1', output_octets='4', update_time='2017-03-10 11:50:00'
```
nemesisdesign: I suppose some of these attributes are shared between the other tests that you created previously, if that's the case please consider creating a dictionary which is reused, eg:

```python
options = _RADACCT.copy()
options['username'] = 'example-of-how-to-change-an-attribute'
RadiusAccounting.objects.create(**options)
```

django_freeradius/tests/test_admin.py
```diff
@@ -213,3 +213,15 @@ def test_nas_admin_save_model(self):
         self.assertNotContains(response, 'error')
         nas.refresh_from_db()
         self.assertEqual(nas.type, 'my-custom-type')
+
+    def test_cleanup_stale_radacct_command(self):
```
nemesisdesign: I only realized now the tests for the management commands shouldn't go in `test_admin.py`. Please move all the tests you created recently to a new file named `test_commands.py`.
You should do this change in a separate commit than the current one but you can leave both in the current PR

docs/source/general/management_commands.rst
```diff
@@ -42,3 +42,18 @@ For example:
 .. code-block:: shell
 
     ./manage.py delete_old_postauth 365
+
+``cleanup_stale_radacct``
+-------------------------
+
+This command closes active RADIUS sessions that are open since <days>.
```
nemesisdesign: Please improve it as follows:

`This command closes stale RADIUS sessions that have remained open for the number of specified `<days>`.`

docs/source/general/management_commands.rst
```diff
@@ -42,3 +42,18 @@ For example:
...
+
+.. code-block:: shell
+
+    ./manage.py cleanup_stale_radacct <days>
+
+For example:
+
+.. code-block:: shell
+
+    ./manage.py cleanup_stale_radacct 30
```
nemesisdesign: let's use 15 because is a more realistic value

django_freeradius/tests/test_admin.py
peppelinux: https://github.com/openwisp/django-freeradius/issues/35

nemesisdesign:
@R9295 tell it to skip tests for the sample app (the app which is needed to test the swappable models), aftr this issue if you could help us to fix #35 it would be really useful because this problem is slowing us down a lot now (we have hit it at least 3-4 times in 1 week), we have GCI task ready for that, @peppelinux just created it - I only have to review it and publish it.
If something is not clear ask us on chat.

R9295:
@nemesisdesign gotcha

PR #83, opened 2017-12-30
[tests] Made tests reusable

R9295:


django_freeradius/tests/base/test_admin.py
```diff
@@ -0,0 +1,194 @@
+from copy import copy
+
+from django.contrib.auth.models import User
+from django.urls import reverse
+
+from django_freeradius import settings
+from django_freeradius.models import (
```
nemesisdesign: the base tests to be reusable by other apps should not import models.

Compares these two files:

base tests: https://github.com/openwisp/django-loci/blob/master/django_loci/tests/base/test_admin.py
actual tests: https://github.com/openwisp/django-loci/blob/master/django_loci/tests/test_admin.py

openwisp-controller imports the base tests of loci to perform an integration test (and adds more tests on top of it):
https://github.com/openwisp/openwisp-controller/blob/master/openwisp_controller/geo/tests/test_admin.py

peppelinux: https://github.com/wq/django-swappable-models

PR #84, opened 2017-12-31
change configuration files path to debian defaults

EdoPut:
this will change all the configuration files path to the one that
ships with debian stretch. I added a note for users with other distro
or version.

Also fixed a sphinx build error with using ~ instead of ^ for
subsubparagraphs

### DIFF
--- a/docs/source/general/freeradius.rst
+++ b/docs/source/general/freeradius.rst
@@ -40,7 +40,13 @@ Now you can install the packages we need:
 Configuring Freeradius 3
 ------------------------
 
-Edit the file ``/etc/freeradius/mods-available/sql``.
+.. note::
+    The path to freeradius configuration could be different on your system. This article use the `/etc/freeradius/3.0/` directory that ships with Debian Stretch
+
+Configure the SQL module
+^^^^^^^^^^^^^^^^^^^^^^^^
+
+Edit the file ``/etc/freeradius/3.0/mods-available/sql``.
 
 Change the configuration for ``driver``, ``dialect``, ``server``, ``port``, ``login``, ``password``, ``radius_db``.
 
@@ -50,7 +56,7 @@ Example configuration using the PostgreSQL database:
 
 .. code-block:: ini
 
-    # /etc/freeradius/mods-available/sql
+    # /etc/freeradius/3.0/mods-available/sql
     driver = "rlm_sql_postgresql"
     dialect = "postgresql"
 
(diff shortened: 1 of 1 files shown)


docs/source/general/freeradius.rst
```diff
@@ -239,7 +242,7 @@ and ``Calling-Station-ID``:
     echo $request | radclient localhost auth testing123
```
peppelinux: radclient for acct and radtest for auth. 
In the comment we read about radtest but example show us a radclient command

EdoPut: Thy are both shown with their relative command

EdoPut:
> We should manage in [...] Debian stable
Debian stretch is the current stable

As for ubuntu there will be a LTS release early in 2018 and we can target that.

I don't understand how a path variable would make the docs better, they are a static page. Maybe we can put some footmarks with a reference to path in other distro.

For what is worth I'm adding a link to the freeradius wiki that explains a lot better that this article the configuration

docs/source/general/freeradius.rst
```diff
@@ -83,15 +87,15 @@ Restart freeradius to load the new configuration:
 
 You may also want to take a look at the `Freeradius documentation <http://freeradius.org/doc/>`_.
 
-How to configure the REST module
---------------------------------
+Configure the REST module
+^^^^^^^^^^^^^^^^^^^^^^^^^
```
nemesisdesign: why this?

nemesisdesign: I mean the use of `^` instead of `-`

EdoPut:
Because is a subsubsection of the subsection "Configure freeradius"

On Dec 31, 2017 19:02, "Federico Capoano" <notifications@github.com> wrote:

*@nemesisdesign* commented on this pull request.
------------------------------

In docs/source/general/freeradius.rst
<https://github.com/openwisp/django-freeradius/pull/84#discussion_r159144883>
:

> @@ -83,15 +87,15 @@ Restart freeradius to load the new configuration:

 You may also want to take a look at the `Freeradius documentation
<http://freeradius.org/doc/>`_.

-How to configure the REST module
---------------------------------
+Configure the REST module
+^^^^^^^^^^^^^^^^^^^^^^^^^

I mean the use of ^ instead of -

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/openwisp/django-freeradius/pull/84#discussion_r159144883>,
or mute the thread

PR #86, opened 2018-01-01
better step by step subdivision

EdoPut:
this will add links to the sql and rest documentation on the freeradius wiki

I also moved to the end of the "Configure Freeradius" section the commands to link files and restart freeradius as it should be done last.


docs/source/general/freeradius.rst
```diff
@@ -162,6 +149,27 @@ For accounting configuration you need to verify that in pre-accounting we have:
         # ...
     }
 
+
```
nemesisdesign: remove this blank line please

PR #87, opened 2018-01-02
[api] check if user is enabled before authorization #63

### LINKED ISSUE #63: [Feature] Disabled Users in API

R9295:
Closes  #63

### DIFF
--- a/django_freeradius/api/views.py
+++ b/django_freeradius/api/views.py
@@ -18,7 +18,7 @@ def authorize(request):
     username = request.data.get('username')
     password = request.data.get('password')
     try:
-        user = User.objects.get(username=username)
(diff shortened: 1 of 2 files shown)


R9295:
@nemesisdesign Travis is stuck at ```npm install -g jslint```

nemesisdesign:
Don't forget to add `closes #<issue>` in your commit messages so issues get closed automatically.

PR #88, opened 2018-01-02
[api] Added accounting start request with missing fields

### LINKED ISSUE #69: [Accounting API] Accounting start fails with coova-chilli
Coova chilli sends an accounting start request which doesn't include the following fields:

- `output_octets`
- `input_octets`
- `session_time`

We should allow these fields to be missing if the request is of type start and default to `0`.

AZtheAsian:
Fixes #69. 

I'm not actually totally sure this solves the problem but I wrote tests for it as well.

### DIFF
--- a/django_freeradius/api/views.py
+++ b/django_freeradius/api/views.py
@@ -63,6 +63,10 @@ class AccountingView(generics.ListCreateAPIView):
         return getattr(self, method)(request, *args, **kwargs)
 
     def create(self, request, *args, **kwargs):
+        is_start = request.data['status_type'] == 'Start'
+        for field in ['session_time', 'input_octets', 'output_octets']:
+            if is_start and request.data[field] == '':
+                request.data[field] = 0
         serializer = self.get_serializer(data=request.data)
         serializer.is_valid()
         error_keys = serializer.errors.keys()
--- a/django_freeradius/tests/base/test_api.py
+++ b/django_freeradius/tests/base/test_api.py
@@ -135,6 +135,43 @@ class BaseTestApi(object):
         self.assertAcctData(ra, data)
 
     @freeze_time(START_DATE)
+    def test_accounting_start_coova_chilli(self):
+        self.assertEqual(self.radius_accounting_model.objects.count(), 0)
+        data = {
+            'status_type': 'Start',
+            'session_id': '5a4f59aa00000001',
+            'unique_id': 'd11a8069e261040d8b01b9135bdb8dc9',
+            'username': 'username',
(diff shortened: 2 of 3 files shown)


AZtheAsian:
@nemesisdesign I changed the serializer, note that I used IntegerField because of the mapping here:
https://github.com/encode/django-rest-framework/blob/master/rest_framework/serializers.py#L841

nemesisdesign:
@AZtheAsian also rebase your fork on the current master please

AZtheAsian:
@nemesisdesign this fixes the tests, but would it fix the issue in an actual use case? The tests emulate an actual post, but we're messing with the views file and not the serializer file

AZtheAsian:
@nemesisdesign I squashed all the commits, I don't understand why we're deleting that first test though

django_freeradius/api/views.py
```diff
@@ -63,6 +63,12 @@ def post(self, request, *args, **kwargs):
         return getattr(self, method)(request, *args, **kwargs)
 
     def create(self, request, *args, **kwargs):
+        if request.data['session_time'] == '' and request.data['status_type'] == 'Start':
```
nemesisdesign: A more elegant implementation which avoids repetition:

```python
is_start = request.data['status_type'] == 'Start'
for field in ['session_time', 'input_octets', 'output_octets']:
    if is_start and request.data[field] == '':
        request.data[field] = 0
```

django_freeradius/tests/base/test_api.py
```diff
@@ -9,13 +9,6 @@
 
 
 class BaseTestApi(object):
-    def test_disabled_user_login(self):
-        User.objects.create_user(username='barbar', password='molly', is_active=False)
-        response = self.client.post(reverse('freeradius:authorize'),
-                                    {'username': 'barbar', 'password': 'molly'})
-        self.assertEqual(response.status_code, 401)
-        self.assertEqual(response.data, {'control:Auth-Type': 'Reject'})
-
```
nemesisdesign: it's tricky to explain.. in short it was a mistake caused by my manual editing and a conflict with your previous patch and another patch from Aarnav, we don't need to delete this test because it's important

django_freeradius/api/views.py
AZtheAsian: that's really helpful, thanks!

sometimes I notice repetition, but can't really find a way to solve it. :/

AZtheAsian:
@nemesisdesign I think travis is broken (it can't connect to ubuntu releases?) so...?

PR #101, opened 2018-03-01
[api] Improved filtering #99

### LINKED ISSUE #99: [Accounting-API] Improve filtering
Allow filtering for these fields:

- `username` filter for a specific username
- `called_station_id`
- `calling_station_id`
- `start_time`: look for sessions started in a specific date (in `YYYY-MM-DD` format)
- `stop_time`: look for sessions finished in a specific date (in `YYYY-MM-DD` format)
- `open`: this is not a database field but a boolean filter that looks for sessions which have `stop_time == None`

I suggest to use this kind of implementation: http://www.django-rest-framework.org/api-guide/filtering/#djangofilterbackend

anurag-ks:
Implements and closes #99 
Also added django-crispy-forms for the DRF’s browsable API.

### DIFF
--- a/django_freeradius/api/views.py
+++ b/django_freeradius/api/views.py
@@ -2,6 +2,8 @@ import drf_link_header_pagination
 import swapper
 from django.contrib.auth import get_user_model
 from django.utils.translation import ugettext_lazy as _
+from django_filters import rest_framework as filters
+from django_filters.rest_framework import DjangoFilterBackend
 from rest_framework import generics
 from rest_framework.decorators import api_view
 from rest_framework.exceptions import ValidationError
@@ -43,6 +45,26 @@ class PostAuthView(generics.CreateAPIView):
 postauth = PostAuthView.as_view()
 
 
+class AccountingFilter(filters.FilterSet):
+    start_time = filters.DateTimeFilter(name='start_time', lookup_expr='gte')
+    stop_time = filters.DateTimeFilter(name='stop_time', lookup_expr='gte')
+    is_open = filters.BooleanFilter(method='filter_open')
+
+    def filter_open(self, queryset, name,  value):
+        if value:
+            return queryset.filter(stop_time__isnull=True)
+        return queryset.filter(stop_time__isnull=False)
+
+    class Meta:
+        model = RadiusAccounting
+        fields = ('username',
+                  'called_station_id',
+                  'calling_station_id',
+                  'start_time',
+                  'stop_time',
+                  'is_open')
+
+
 class AccountingViewPagination(drf_link_header_pagination.LinkHeaderPagination):
     page_size = 10
     page_size_query_param = 'page_size'
@@ -63,6 +85,8 @@ class AccountingView(generics.ListCreateAPIView):
     queryset = RadiusAccounting.objects.all()
     serializer_class = RadiusAccountingSerializer
     pagination_class = AccountingViewPagination
+    filter_backends = (DjangoFilterBackend,)
+    filter_class = AccountingFilter
 
     def post(self, request, *args, **kwargs):
         status_type = self._get_status_type(request)
--- a/django_freeradius/tests/base/test_api.py
+++ b/django_freeradius/tests/base/test_api.py
@@ -313,6 +313,102 @@ class BaseTestApi(object):
         self.assertEqual(item['nas_ip_address'], '172.16.64.91')
         self.assertEqual(item['calling_station_id'], '5c:7d:c1:72:a7:3b')
 
+    def test_accounting_filter_username(self):
+        data1 = self.acct_post_data
+        data1.update(dict(username='test_user',
+                          unique_id='75058e50'))
+        self.radius_accounting_model.objects.create(**data1)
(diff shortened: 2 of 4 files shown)


anurag-ks:
Should I write new tests for this or modify the old ones ?

requirements-test.txt
```diff
@@ -10,3 +10,4 @@ sphinx
 sphinx_rtd_theme
 freezegun
 openwisp-utils>=0.2,<0.3
+django-crispy-forms
```
nemesisdesign: bind its requirements as with `openwisp-utils`

django_freeradius/api/views.py
```diff
@@ -43,6 +45,26 @@ def post(self, request, *args, **kwargs):
 postauth = PostAuthView.as_view()
 
 
+class AccountingFilter(filters.FilterSet):
+    start_time = filters.IsoDateTimeFilter(name='start_time')
+    stop_time = filters.IsoDateTimeFilter(name='stop_time')
+    open = filters.BooleanFilter(label='Open', method='filter_open')
+
+    def filter_open(self, queryset, name,  value):
+        if(value):
```
nemesisdesign: `if value:` (parentheses are not needed)

django_freeradius/api/views.py
```diff
@@ -43,6 +45,26 @@ def post(self, request, *args, **kwargs):
 postauth = PostAuthView.as_view()
 
 
+class AccountingFilter(filters.FilterSet):
+    start_time = filters.IsoDateTimeFilter(name='start_time')
+    stop_time = filters.IsoDateTimeFilter(name='stop_time')
+    open = filters.BooleanFilter(label='Open', method='filter_open')
```
nemesisdesign: let's change this to `is_open` to avoid overriding the `open` reserved keyword, only noticed this now.

Remove the `label` if you can for consistency, if we add it we should add it on all the other fields

requirements-test.txt
nemesisdesign: I sent you a few messages in the public chat regarding this, if not required I suggest to remove it because doesn't seeem to bring much value and will slow down the build by increasing the build time

django_freeradius/tests/base/test_api.py
```diff
@@ -313,6 +313,61 @@ def test_accounting_list_200(self):
...
+        self.assertEqual(len(response.json()), 1)
+        self.assertEqual(response.status_code, 200)
+        item = response.data[0]
+        self.assertEqual(item['called_station_id'], 'E0-CA-40-EE-D1-0D')
+        response = self.client.get('{0}?calling_station_id=4c:8d:c2:80:a7:4c'.format(self._acct_url))
+        self.assertEqual(len(response.json()), 1)
+        self.assertEqual(response.status_code, 200)
+        item = response.data[0]
+        self.assertEqual(item['calling_station_id'], '4c:8d:c2:80:a7:4c')
+        response = self.client.get('{0}?start_time=2018-03-02T00:43:24.020460+01:00'.format(self._acct_url))
```
nemesisdesign: we should allow users (developers) to filter by a date and not a datetime.
That is, we want to filter by the entire date, eg: `2018-03-02` and not `2018-03-02T00:43:24.020460+01:00`

But since you will be freezing the date on top with `@freeze_time`, use `START_DATE` as follows:
```python
response = self.client.get('{0}?start_time={1}'.format(self._acct_url, START_DATE))
```
I just tried it and it seems to pass this assertion.

django_freeradius/tests/base/test_api.py
```diff
@@ -313,6 +313,61 @@ def test_accounting_list_200(self):
         self.assertEqual(item['nas_ip_address'], '172.16.64.91')
         self.assertEqual(item['calling_station_id'], '5c:7d:c1:72:a7:3b')
 
+    def test_accounting_filter_200(self):
```
nemesisdesign: use `@freeze_time(START_DATE)` as in other tests (look up on google to find out what this library does)

tests/settings.py
```diff
@@ -28,6 +28,8 @@
     'django.contrib.admin',
     'django_freeradius',
     'rest_framework',
+    'django_filters',
+    'crispy_forms',
```
nemesisdesign: I sent you a few messages in the public chat regarding this, if not required I suggest to remove it because doesn't seeem to bring much value and will slow down the build by increasing the build time

django_freeradius/api/views.py
anurag-ks: Yeah I was going to tell you that.

anurag-ks:
@nemesisdesign I added django-crispy-forms for this - 
screenshot-2018-3-4 accounting django rest framework 1

nemesisdesign:
@anurag-ks PR is perfect as it's now without the additional dependency, will repeat what I've said in chat just for the public records: we want to keep the automated test environment minimal in order to not slow down the build. OpenWISP follows the unix philosophy of minimal software which is extensible.
That means developers who want to add extensions in their development environment can do so without any problem and without committing those in the repo.

Less is more :-)

PR #106, opened 2018-03-09
[db] Altered migration files to improve freeradius3 support

### LINKED ISSUE #41: [db] Smart migration from freeradius 3
As discussed, we want to make it easy for those users who are already using freeradius 3 to start using django-freeradius.

To accomplish this we should proceed as follows:

1. [ ] first of all, we have to ensure our models reflect the default freeradius schema very closely
2. [ ] the first django migration should be dedicated to the default freeradius schema
3. [ ] every modification and addition to the default schema, must be done in subsequent migrations, starting from `0002`
4. [ ] we then need to document how to skip the first migration in case users want to use django-freeradius on an existing freeradius 3 database

rohithasrk:
Closes #41


nemesisdesign:
@rohithasrk I'm delighted to see this. Could you describe the process with which you tested this and ensured it works as expected?

I'll do my own tests soon but I first need to know how you did yours.

rohithasrk:
@nemesisdesign. So basically, I marked all the databases and fields present in the default freeradius3 schema by checking out the `/mods-config/sql/main/postgresql/schema.sql` and also the documentation/ blogs on freeradius. I see that the models `RadiusGroup` and `RadiusGroupUsers` aren't there in freeradius by default. Therefore, moved them to the second migration file and the additional files in `RadiusCheck` models to third. The last migration file is the indexing of the `called-station-id` and `calling-station-id` fields.

For ensuring that it works I've made a new database and migrated the first migration and checked some basic functionalities with the sql module enabled in accounting and authorization. They seem to work well. I'm curious if there is any better and automated way to test it.

nemesisdesign:
Great to know that @rohithasrk.

Now that `RadiusGroup` and `RadiusGroupUsers` have been removed we can remove them from here too.

There's one thing to keep in mind: changing existing migrations will break migrations for users that are using this (unreleased) version. 
Even though this project is still unreleased, it seems some users are already using it and some of them are contributing.. so it wouldn't be nice if we break their possibility to upgrade.

The migration framework has a concept called "squashing migrations": https://docs.djangoproject.com/en/2.0/topics/migrations/#migration-squashing
This should allow us to create a couple of squashed migrations, leave the old migrations for old users to upgrade, urge these users to upgrade, then when we are done with this process we can delete the old unsquashed migrations.

What do you think? I don't want to spend too much time on this issue but I also don't want to break the system of our contributors.

rohithasrk:
@nemesisdesign Should we have the old migrations in a separate directory or so? Because having them in the same directory might cause errors while applying migrations.

nemesisdesign:
@rohithasrk I'm not an expert when it comes to squashing migrations.. please refer to the advice given in the django docs, not sure if these can be stored in a separate directory

PR #107, opened 2018-03-11
[docs] Documented API usage

### LINKED ISSUE #102: [docs] Document API usage
Add an API page which explains briefly what each end point does, we can use `.. automethod::` and or `.. autofunction:: ` as we do in the [netjsonconfig docs](http://netjsonconfig.openwisp.org/en/stable/general/basics.html#implementation-details) to automatically include the docstrings of the views and link their source code.

Regarding the accounting API, we should very briefly describe how to use the new features just added in #101:
- pagination & header pagination
- filtering

anurag-ks:
Implements and closes #102

Use of automethod or autofunction didn't seem appropriate because we are going to document API endpoints and not functions, right ?
@nemesisdesign @rohithasrk please review it and suggest if any more content should be added or not. ?

### DIFF
--- a/docs/source/general/api.rst
+++ b/docs/source/general/api.rst
@@ -1,0 +1,158 @@
+===================
+API Documentation
+===================
+
+
+
+Accounting
+###########
+
+.. code-block:: text
+
+    /api/accounting
+
+GET
+++++++
+Returns a list of accounting objects
+
+.. code-block:: text
+
+    GET /api/accounting
+
+.. code-block:: json
+
+    [
+      {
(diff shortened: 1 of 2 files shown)


anurag-ks:
@nemesisdesign I felt that reading separate sections rather than reading single long page while scrolling down, would be easier. Also if the API becomes big and complex in the future we might as well have to create new pages for them, here again I am not sure about it, this is just a suggestion.

anurag-ks:
@nemesisdesign made it into a single page :+1:

nemesisdesign:
thx :+1:

Issue #39, opened 2017-07-27
[authorize] Allow to enforce session limits


nemesisdesign:
We have to mainly 2 things:

- [ ] add a way (or establish what's the best way using the current available SQL tables) to define limits to the session of the users
- [ ] enforce these limits in the authorize API method

More specifically, we need to be able to replicate at least two features of [OpenWISP User Management System](https://github.com/openwisp/OpenWISP-User-Management-System):

- [ ] daily bandwidth counter
- [ ] daily session limit

We still need to understand in practice what's the best way to implement this feature.

rohithasrk:
Working on this.

PR #110, opened 2018-03-13
[feature] Enforce session and bandwidth limits

rohithasrk:
A working model of #39. This implementation requires rlm_sqlcounter(for session limits) and some extra configurations for bandwidth usage to be configured with freeradius. The link to my freeradius config could be found out [here](https://github.com/rohithasrk/raddb). The data-bandwidth feature hasn't been tested properly yet. Things to do:

- [x] Document and add instructions
- [x] Unit tests for each of these
- [x] Test the data-bandwidth feature more exhaustively although this is how we shall implement it.


nemesisdesign:
There's a way to implement this feature without additional modules, read here: https://wiki.freeradius.org/modules/Rlm_sqlcounter

Look for additional information regarding `sqlcounter`.

nemesisdesign:
@rohithasrk these 2 models: https://github.com/openwisp/django-freeradius/blob/master/django_freeradius/base/models.py#L115-L164 are mirroring how the functionality was implemented in [OpenWISP User Management system](https://github.com/openwisp/OpenWISP-User-Management-System) (from now OWUMS).

Most deployments of OWUMS had 2 default groups:

- standard users (limited usually to a certain limit of hours and/or bandwidth)
- power users (unlimited)

A user could be member of more groups and the priority field decided which rules applied.
The checks and replies where somehow related to these 2 models, but I don't remember how exactly.

We need to dig into this more, here's a few links to the old implementation:

- RadiusGroup model: https://github.com/openwisp/OpenWISP-User-Management-System/blob/master/app/models/radius_group.rb
- migrations for radius group
  - https://github.com/openwisp/OpenWISP-User-Management-System/blob/master/db/migrate/20091228102900_create_radius_groups.rb
  - https://github.com/openwisp/OpenWISP-User-Management-System/blob/master/db/migrate/20091228103337_create_radius_groups_users.rb

Note: we don't need to achieve the same level of flexibility but we don't have to follow exactly the same way of implementing it, we can introduce improvements. Just keep in mind radius checks may represent other things than bandwidth and time limit so we need to provide a flexible way to define these (eg: adding fields like "session limit" or "bandwidth" would limit to these use cases only).

nemesisdesign:
So to summarize, our use case is bandwidth and session limit but in the truth the feature aims at creating groups of radius checks and radius replies that can be assigned to users, so administrators can implement different rules for different groups like "guests", "regular visitor", "power user", ecc.

django_freeradius/base/models.py
```diff
@@ -91,13 +92,16 @@
...
                      ('+=', '+='))
 
-RADCHECK_PASSWD_TYPE = ['Cleartext-Password',
-                        'NT-Password',
-                        'LM-Password',
-                        'MD5-Password',
-                        'SMD5-Password',
-                        'SSHA-Password',
-                        'Crypt-Password']
+RADCHECK_ATTRIBUTE_TYPES = ['Cleartext-Password',
```
nemesisdesign: have you tested how this plays with the password hashing feature that is also being changed in #131?

django_freeradius/base/models.py
```diff
@@ -652,3 +656,77 @@ def expire(self):
...
+        verbose_name_plural = _('radius session and bandwidth limits')
+        abstract = True
+
+    def __str__(self):
+        return self.name
+
+    def save(self):
+        if self.default:
+            RadiusProfile = swapper.load_model('django_freeradius', 'RadiusProfile')
+            for rl in RadiusProfile.objects.filter(default=True):
```
nemesisdesign: I think you can optimize this line, should be (writing by memory):

```python
RadiusProfile.objects.filter(default=True).update(default=False)
```

This will generate a single update query (but won't call the `save()` method of the model, which should be ok in this case.

django_freeradius/base/models.py
2stacks: Let me know how I can help.  Password is definitely only one of many radcheck attribute types.

nemesisdesign: @2stacks yes, that would be great, you could help us by testing how this patch interacts with yours. You should wait until we are ready to merge here before doing that. Hopefully this means tomorrow if @rohithasrk can make it.

rohithasrk:
@nemesisdesign The things remaining to do are, adding the field as an inline field in User. My question here is, do we inherit the user model and do it? As because while implementing multitenancy we anyways would use the models from openwisp_users. So right now, I'm confused about how we go about implementing the inlines and make them reusable as well.

Also added the default migration. Need to add the configurable parameters. 

The SAMPLE_APP issue is still left unsolved. Will do it today.

django_freeradius/migrations/0018_add_default_profile.py
```diff
@@ -0,0 +1,41 @@
...
+
+from django.conf import settings
+from django.contrib.auth import get_user_model
+from django.db import migrations
+
+def add_default_profiles(apps, schema_editor):
+    RadiusProfile = swapper.load_model('django_freeradius', 'RadiusProfile')
+    default_profile = RadiusProfile.objects.filter(default=True)
+    if not default_profile.exists():
+        limited_user = RadiusProfile(name="Limited User", default=True)
```
nemesisdesign: excuse me, is this profile doing anything at all? Where are the limits defined here?

django_freeradius/migrations/0018_add_default_profile.py
```diff
@@ -0,0 +1,41 @@
...
+        limited_user.save()
+        power_user = RadiusProfile(name="Power User")
```
nemesisdesign: same question here

django_freeradius/migrations/0018_add_default_profile.py
rohithasrk: Nothing as of now. Wanted to confirm the limits from you.

rohithasrk: The parameters are what I wanted to confirm? What limits do we put on each of them

django_freeradius/migrations/0018_add_default_profile.py
```diff
@@ -0,0 +1,45 @@
...
+def add_default_profiles(apps, schema_editor):
+    RadiusProfile = swapper.load_model('django_freeradius', 'RadiusProfile')
+    default_profile = RadiusProfile.objects.filter(default=True)
+    if not default_profile.exists():
+        limited_user = RadiusProfile(name="Limited User", default=True,
+                                     daily_session_limit=10800,
+                                     daily_bandwidth_limit=3000000000)
+        limited_user.save()
+        power_user = RadiusProfile(name="Power User",
+                                   daily_session_limit=86400,
```
nemesisdesign: double this just in case we ever get a user that stays connected 24 hours straight and the NAS doesn't calculate the time perfectly and may log them out

django_freeradius/base/admin.py
```diff
@@ -206,3 +208,21 @@ def get_readonly_fields(self, request, obj=None):
 
 
 AbstractRadiusBatchAdmin.list_display += ('expiration_date', 'created')
+
+
+class AbstractRadiusProfileAdmin(TimeStampedEditableAdmin):
+    pass
+
+
+class RadiusUserProfileInline(StackedInline):
```
nemesisdesign: add this mixin from `openwisp-utils`: https://github.com/openwisp/openwisp-utils/blob/master/openwisp_utils/admin.py#L93

rohithasrk:
Great @nemesisdesign! Thank you :smile:

PR #109, opened 2018-03-12
[feature] Import and generate users using CSV and prefixes

rohithasrk:
A basic implementation of the batch add users feature. As of now, it is just a management command which does it. ToDos:

- [x] Add from the admin interface
- [x] Write unit tests for all the added code
- [x] Discuss the format of the CSV file and edit the handle function accordingly.
- [x] Write documentation that explains how to import users from an existing database
- [x] Write documentation that explains how to create temporary users for events

Implementation: Created a new model by the name RadiusBatch having m2m relationship with the RadiusCheck model and a parameter named expiration_date(so that it could be used in future) and added a management command to add users from a csv file. I assumed the csv fomat to be
```
    <username> <password>
```
for now, which could be easily changed. @nemesisdesign Let me know if the method could be improved in any way.

PR #117, opened 2018-05-23
[WIP] Add users from admin interface

rohithasrk:
This PR is an extension of #109 which aims to add users from the admin interface. Things to do:

- [ ] Handle the csv upload command by calling batch model functions.
- [ ] Decide whether to store the csv file for not.
- [ ] Tests and docs for the same.


rohithasrk:
Merged with #109

Issue #128, opened 2018-07-01
[bug] Not all password hashers are implemented for radcheck model


2stacks:
Currently only 'Cleartext-Password', 'LM-Password' and 'NT-Password' have been implemented.  Adding an entry to the 'Radius Checks' Admin form with any of the following Attributes;

MD5-Password
SMD5-Password
SSHA-Password
Crypt-Password

Results in the following error;

Exception Type:  UnboundLocalError
Exception Value:  local variable 'password_renewed' referenced before assignment

PR #129, opened 2018-07-01
[tests] Added tests for #128

2stacks:
Added tests for all password hasher attribute types available in
'Radius Checks' admin form.

Tests #128


2stacks:
I'll work on a fix as well.  I was hoping to have a discussion regarding which hashers should be implemented.  The freeradius documentation only covers the currently mentioned hashing techniques however, I know from practice that using more advanced hashes (ie. SSHA512) will work with freeradius and certain NAS types.  I'll stick to the current list mentioned in #128 but would like to add support for others at some point.

PR #130, opened 2018-07-03
[api] add post calls for RadiusBatch

AlessiaVe:
Add two calls for create with the Rest Api a RadiusBatch, one with the csv and one with the prefix.
I created CreateAPIViews with the following url: /api/v1/batchCsv and /api/v1/batchPrefix.


django_freeradius/api/views.py
```diff
@@ -137,3 +139,81 @@ def _get_status_type(self, request):
...
+            batch = RadiusBatch(name=name, strategy=strategy, expiration_date=expiration_date, prefix=prefix)
+            number_of_users = self._get_number_of_users(request)
+            batch.prefix_add(prefix, number_of_users)
+            response = RadiusBatchPrefixSerializer(batch)
+            return Response(response.data, status=status.HTTP_201_CREATED)
+        return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
+
+    def _get_strategy(self, request):
+        try:
+            s = request.data['strategy']
```
nemesisdesign: please use readable variable names, change `s` to `strategy`, keep in mind code is read a lot more than is written (and the one reading this code over and over will be more likely me :-P)

django_freeradius/api/views.py
```diff
@@ -137,3 +139,81 @@ def _get_status_type(self, request):
...
+    serializer_class = RadiusBatchPrefixSerializer
+
+    def post(self, request, *args, **kwargs):
+        serializer = RadiusBatchPrefixSerializer(data=request.data)
+        if serializer.is_valid():
+            name = serializer.data.get('name')
+            strategy = self._get_strategy(request)
+            expiration_date = serializer.data.get('expiration_data')
+            prefix = serializer.data.get('prefix')
+            batch = RadiusBatch(name=name, strategy=strategy, expiration_date=expiration_date, prefix=prefix)
```
nemesisdesign: are you sure you need to create this instance here?
the serializer should be able to save the object and return the instance, please check this, we want to use DRF properly and take advantage of its well tested and expected behaviour

django_freeradius/api/urls.py
```diff
@@ -6,4 +6,6 @@
     url(r'^authorize/$', views.authorize, name='authorize'),
     url(r'^postauth/$', views.postauth, name='postauth'),
     url(r'^accounting/$', views.accounting, name='accounting'),
+    url(r'^v1/batchCsv/$', views.batchCsv, name='batchCsv'),
```
nemesisdesign: adding `v1` here is not the right thing to do, we'll have to add it one level app.
also, the camelCase convention is not suited here, therefore please change this line to:

```python
url(r'^batch-csv/$', views.batchCsv, name='batch-csv'),
```

django_freeradius/api/urls.py
```diff
@@ -6,4 +6,6 @@
...
+    url(r'^v1/batchPrefix/$', views.batchPrefix, name='batchPrefix'),
```
nemesisdesign: same as previous line, change to 
```python
url(r'^batch-prefix/$', views.batchPrefix, name='batch-prefix'),
```

django_freeradius/api/serializers.py
```diff
@@ -63,3 +65,53 @@ def validate(self, data):
...
+class UserSerializer(serializers.ModelSerializer):
+    groups = GroupSerializer(many=True)
+
+    class Meta:
+        model = User
+        fields = '__all__'
+
+
+class RadiusBatchCsvSerializer(serializers.ModelSerializer):
+    name = serializers.CharField()
```
nemesisdesign: are you sure this is needed? Try to remove it, the `ModelSerializer` class should be able to determine this field automatically.
Same applies to all the other model fields for which you don't need to do customizations

django_freeradius/api/serializers.py
```diff
@@ -63,3 +65,53 @@ def validate(self, data):
...
+    prefix = serializers.CharField(required=False, read_only=True)
+    expiration_date = serializers.DateField(required=False)
+    pdf = serializers.FileField(required=False, read_only=True)
+
+    class Meta:
+        model = RadiusBatch
+        fields = '__all__'
+
+    def create(self, validated_data):
+        return RadiusBatch.objects.create(**validated_data)
```
nemesisdesign: why is this needed?

django_freeradius/api/serializers.py
```diff
@@ -63,3 +65,53 @@ def validate(self, data):
...
+
+
+class RadiusBatchPrefixSerializer(serializers.ModelSerializer):
```
nemesisdesign: there's a lot of duplication with the other serializer, can you define the common fields in a shared class?

django_freeradius/api/views.py
```diff
@@ -75,9 +79,7 @@ class AccountingView(generics.ListCreateAPIView):
     """
     HEADER: Pagination is provided using a Link header
             https://developer.github.com/v3/guides/traversing-with-pagination/
-
     GET: get list of accounting objects
-
```
nemesisdesign: don't touch other parts of the code that are not related to your patch please, it's a bad practice

django_freeradius/api/views.py
AlessiaVe: I did this because I want to create the users associated to the RadiusBatch, and I can't do this without this instance. I will check if I can do this without the instance but this is the only way that I found.

django_freeradius/api/serializers.py
AlessiaVe: it's an error, I delete it this evening in the new commit! Sorry!

AlessiaVe:
@nemesisdesign I made one call for the two calls and made all the tests. The code now is  very shorter. I try to do the call without the instance of the RadiuBatch but if I delete it the call doesn't work! Tell me if I must do other changes!

django_freeradius/tests/base/test_api.py
```diff
@@ -331,6 +332,7 @@ def test_accounting_400_invalid_status_type(self):
     def test_accounting_400_validation_error(self):
         data = self.acct_post_data
         data['status_type'] = 'Start'
+
```
nemesisdesign: please avoid adding this blank line which is unrelated to your patch

docs/source/general/api.rst
```diff
@@ -177,3 +177,37 @@ Sets the response data to None in order to instruct
...
 Responds only to **POST**
+
+Batch
+########
+
+.. code-block:: text
+
+    /api/batch
+
+Responds to only **POST**, used to save a RadiuBatch instance.
```
nemesisdesign: should be `Responds only to`

nemesisdesign: use
```
``RadiuBatch``
```
so it will be styled as code

docs/source/general/api.rst
```diff
@@ -177,3 +177,37 @@ Sets the response data to None in order to instruct
...
+It returns the information of the batch operation and the list of the users generated.
+It is possible to generate the users of the RadiusBatch with two different strategies: csv or prefix.
```
nemesisdesign: use
```
``RadiuBatch``
```
so it will be styled as code

django_freeradius/api/views.py
```diff
@@ -161,3 +162,44 @@ def _get_status_type(self, request):
...
+                batch.prefix_add(prefix, number_of_users_validated)
+                response = RadiusBatchSerializer(batch)
+            return Response(response.data, status=status.HTTP_201_CREATED)
+        return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
+
+    def _validate_number_of_users(self, number):
+        if number > 0:
+            return number
+        else:
+            raise ValidationError({'number_of_users': [_('Number of users should be greater than 0.')]})
```
nemesisdesign: I think you can accomplish this much more easily because it's a feature already provided by django-rest-framework for integer fields, see the `min_value` argument of the serializer `IntegerField` here: http://www.django-rest-framework.org/api-guide/fields/#integerfield

PR #131, opened 2018-07-03
[models] Implemented default password hashers #128

### LINKED ISSUE #128: [bug] Not all password hashers are implemented for radcheck model

2stacks:
I've added code to hash passwords as suggested in http://www.openldap.org/faq/data/cache/347.html. 
There is more than one way to go about this but since the intent is to use django's auth_user table this effort only benefits anyone not using the Rest API for authentication.  None of these password hashing methods should be used.  Freeradius can support more advanced hashing methods as described in https://freeradius.org/radiusd/man/rlm_pap.html. Many of these are OS dependent and compatibility with LDAP may require RFC 2307 style headers.  Perhaps a set of supported use cases can be defined, for using radcheck for authentication instead of or in addition to the auth_user table, so that more secure hashing methods can be implemented.

 Fixes #128


nemesisdesign:
@peppelinux since you added this feature, could you help out in testing and reviewing this patch?

2stacks:
For anyone interested in testing, make sure freeradius is only configured with sql (not rest) in the authorize section.  All password hashers were tested using radtest on both debian stretch and alpine linux 3.7.

2stacks:
Im not sure I understand what you mean.

rohithasrk:
Yes basically. Because this function is going to break if an attribute other than the above ones are specified. Like if I put in a new non password attribute like Max-Session-Limit, it probably doesn't need any encoding or hashing.

2stacks:
I think we can avoid this by adding more if statements to the model manager?

2stacks:
In my commit I remove the model field blank=True.  Is there a use case for
this?  I cant think of any entry in the radcheck table that is usable
without an attribute type.

rohithasrk:
I don't think so. But you might have to generate new migration files.

2stacks:
Agreed, this was committed in #131

2stacks:
I've added the requested changes.  I think this should merge with #110 now.  We still need to address issue #134. I think the admin form should  keep the existing field restrictions if the attribute is a password. I recommend we differentiate between password and non password attributes.  Also, it doesn't hurt but it's confusing to run non password attributes through the _encode_secret function.

2stacks:
I rebased this pull request on the recent merge of #110.  Please review and let me know if it is acceptable.

2stacks:
No problem.  Should I rebase again with latest merges to master?

2stacks:
Id be more than happy to add documentation.  Let me know if you want it as
part of this PR.  I have a busy week so it may not be before the weekend.

2stacks:
I'll rebase again.  The #138 merge has conflicts with the tests I added.

rohithasrk:
Yeah @2stacks. Basically, we are reusing the tests in the sample_app and not duplicating. So we'd just need to import them and make sure they work here.

2stacks:
No problem, I'll rebase and modify my tests to match.

rohithasrk:
Give me 5 more minutes as I'm going to push one more commit to master.

2stacks:
Im working on some now.  Do you have a preference for page title?  This
feature covers both Authorization and Authentication in Freeradius.

nemesisdesign:
@2stacks "Authorization with Radius Checks"? Anything better in mind?

2stacks:
Nope, that's along the lines of what I was thinking.

On Tue, Jul 10, 2018 at 10:08 AM, Federico Capoano <notifications@github.com

2stacks:
Just one more question, should this documentation go on its own page or be
a subsection under "Installation and configuration of Freeradius3" section?

PR #132, opened 2018-07-03
[radcheck.js] Don't hide newValueRow if value not set #127

### LINKED ISSUE #127: Show the value field in admin panel on the attribute dropdown change
As of now to add a radcheck instance from the admin panel we are having to do a double work as the value field isn't displayed in the add view. I think we'll need to display it.

2stacks:
This can be improved by adding an option to mask/hide the existing 
hash/password if set.

Fixes #127


nemesisdesign:
Thanks @2stacks

PR #135, opened 2018-07-05
reused tests in sample_radius/tests.py. Fixes #119

### LINKED ISSUE #119: [tests] Are we really reusing base tests in the sample app?
Recently we worked on #83, but checking the tests in the sample app it doesn't seem to me that [the base tests are being imported](https://github.com/openwisp/django-freeradius/blob/1d2f7d8890d331ceb322d561c4a3e3649e6bb9ed/tests/sample_radius/tests.py) and hence I'm not sure the base tests are run in the sample app too.

R9295:


R9295:
Just went through that right after the build failed, it seems that a couple of new tests were added in base/test_admin.py since I last checked. So I went ahead and removed the inheritance.

nemesisdesign:
@R9295 sorry only now I understood what you said about the removed inheritance, what can we do to add back the inheritance and remove the duplicated test logic?

nemesisdesign:
@R9295 patch looks good, try rebasing your branch on the current master and update the PR, let's see if everything is still fine

PR #138, opened 2018-07-07
reused tests in sample_radius/tests.py.

### LINKED ISSUE #119: [tests] Are we really reusing base tests in the sample app?

R9295:
Fixes #119 
Sorry for the resend

### DIFF
--- a/django_freeradius/management/commands/batch_add_users.py
+++ b/django_freeradius/management/commands/batch_add_users.py
@@ -1,12 +1,14 @@
 import sys
 from datetime import datetime
 
+import swapper
 from django.core.exceptions import ValidationError
 from django.core.files import File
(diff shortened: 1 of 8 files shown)


R9295:
Yes

R9295:
Whoops, will do

rohithasrk:
Hey @R9295. Also check out https://coveralls.io/jobs/38181512. I think you are missing out the management commands too.

R9295:
Added the test

PR #140, opened 2018-07-09
[tests] Added the missing tests to sample_radius

rohithasrk:
Improves #138


rohithasrk:
While going through coveralls and the code, I still realised that we were not using all tests in sample_app yet. Also, notices that some of the tests are being repeated so added a skipIf decorator on them. This app is a great example of reusing all the tests and models.

nemesisdesign:
:+1:

PR #144, opened 2018-07-12
Token based authentication for the API

### LINKED ISSUE #33: [API] How to limit usage?
What's the best and simplest way to limit who can use the API?

Initially I thought about allowing specific ip ranges but then I thought maybe using some sort of token authentication may be good too.

PS: we should choose a solution which is also simple to automate via ansible

rohithasrk:
Implements and closes #33

### DIFF
--- a/django_freeradius/api/views.py
+++ b/django_freeradius/api/views.py
@@ -1,13 +1,17 @@
 import drf_link_header_pagination
 import swapper
 from django.contrib.auth import get_user_model
+from django.contrib.auth.models import AnonymousUser
 from django.utils.translation import ugettext_lazy as _
 from django_filters import rest_framework as filters
 from django_filters.rest_framework import DjangoFilterBackend
 from rest_framework import generics
-from rest_framework.decorators import api_view
-from rest_framework.exceptions import ValidationError
+from rest_framework.authentication import BaseAuthentication
+from rest_framework.exceptions import AuthenticationFailed, ValidationError
 from rest_framework.response import Response
+from rest_framework.views import APIView
+
+from django_freeradius.settings import API_TOKEN
 
 from .serializers import RadiusAccountingSerializer, RadiusPostAuthSerializer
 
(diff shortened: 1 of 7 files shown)


django_freeradius/api/views.py
```diff
@@ -16,27 +18,47 @@
...
+        if request.META.get('HTTP_AUTHORIZATION', False):
+            headers = request.META.get('HTTP_AUTHORIZATION').split(',')
+            for header in headers:
+                token = header.split(' ')[1]
+                if token == API_TOKEN:
+                    return True
+        return False
+
+
+class AuthorizeView(TokenAuthMixin, APIView):
```
nemesisdesign: Good to see this being converted to a class based view, but I don't understand.. why do we want to reply with the radius terminology?
Someone is trying to use our API even though we haven't authorized them, we can just cut them off, we don't care if freeradius will complain because of some error, we can just document this like: "if you don't configure freeradius to use the token you will get this error...".

I believe you should create a custom DRF authentication class which returns an anonymous user if successful or raises an exception if unsuccessful: http://www.django-rest-framework.org/api-guide/authentication/#custom-authentication.
You may also need to set the permission attribute as well.

rohithasrk: @nemesisdesign Yeah I just left kept it for testing purposes. We could change the reply here.

Custom authentication sounds good actually. Will modify accordingly.

docs/source/general/api.rst
```diff
@@ -2,6 +2,24 @@
...
+Set this variable to the secret API token of your instance.
+Users can send this token as a bearer token or also as a query string.
+FreeRADIUS must be configured to send a bearer token as mentioned in the
+`docs <freeradius.html>`_.
+
+Accessing the API
+-----------------
+
+To access the API you will need to send the token as a querystring or a
+bearer token in the Authorization header.
```
nemesisdesign: Please mention the two options in a list, put the bearer token option first and write `(recommended)` besides it.

Then please provide 2 examples with `curl`, one with the bearer token header and one with the querystring.

django_freeradius/settings.py
```diff
@@ -35,3 +35,5 @@
                                 os.path.dirname(__file__),
                                 'templates/django_freeradius/prefix_pdf.html')
                              )
+
+API_TOKEN = getattr(settings, 'DJANGO_FREERADIUS_API_TOKEN', None)
```
nemesisdesign: oh sorry. I just realized this! What happens if the app is initialized with no token?
I think it the safest thing is to add a `ready` method here: https://github.com/openwisp/django-freeradius/blob/master/django_freeradius/apps.py
and raise `ImproperlyConfigured` if the token is not a string of at least 15 characters long

rohithasrk: Fixed.

django_freeradius/api/views.py
```diff
@@ -16,20 +20,40 @@
...
+class TokenAuthentication(BaseAuthentication):
+    def authenticate(self, request):
+        if request.META.get('HTTP_AUTHORIZATION', False):
+            headers = request.META.get('HTTP_AUTHORIZATION').split(',')
+            for header in headers:
+                token = header.split(' ')[1]
+                if token == API_TOKEN:
+                    return (AnonymousUser(), None)
+        if request.GET.get('token', False):
+            if request.GET.get('token') == API_TOKEN:
```
nemesisdesign: can we simplify these two nested `if` by doing simply:

```python
if request.GET.get('token') == API_TOKEN:
    # ...
```

If the token is not set the logic will never get here because we are raising `ImproperlyConfigured`.

django_freeradius/apps.py
```diff
@@ -1,6 +1,17 @@
...
 
 
 class DjangoFreeradiusConfig(AppConfig):
     name = 'django_freeradius'
     verbose_name = 'Freeradius'
+
+    def check_settings(self):
+        if API_TOKEN and len(API_TOKEN) < 15 or not API_TOKEN:    # pragma: nocover
+            raise ImproperlyConfigured('Security error: API_TOKEN is either not set or '
+                                       'is less than 15 characters.')
```
nemesisdesign: the error message should mention the public name of the setting, `DJANGO_FREERADIUS_API_TOKEN`, the user won't be sure of what `API_TOKEN` is.

PR #145, opened 2018-07-12
[docs] Document using Radius Checks for authorization

2stacks:
This commit documents Django-Freeradius' support for Freeradius 
compatible password hashing and storage.

Documents #128

### DIFF
--- a/docs/source/general/freeradius.rst
+++ b/docs/source/general/freeradius.rst
@@ -71,6 +71,7 @@ Example configuration using the PostgreSQL database:
     password = "<password>"
     radius_db = "radius"
 
+.. _configure-rest-module:
 
 Configure the REST module
 ^^^^^^^^^^^^^^^^^^^^^^^^^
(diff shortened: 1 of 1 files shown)


2stacks:
@nemesisdesign @rohithasrk I've submitted a draft for your review.  I tried to be verbose so that you can edit it down to what you feel is important.  If I have spoken in error on behalf of the goals/intent of the project please forgive me.

docs/source/general/freeradius.rst
```diff
@@ -208,6 +209,76 @@ An example using MySQL is:
...
+
+To configure support for accessing user credentials with Radius Checks edit the ``authorize`` section
+of the ``/etc/freeradius/3.0/sites-available/default`` configuration file as follows
+
+.. code-block:: ini
+
+    # /etc/freeradius/3.0/sites-available/default
+
+    authorize {
+       #rest
```
nemesisdesign: one question: can we implement a way to use both? If yes do you have any idea how?
@2stacks @peppelinux

2stacks: I believe so.  I mentioned somewhere earlier that the reason it currently fails is that the Rest API returns Auth-Type := Accept and/or Reject.  Doing this prevents Freeradius from being able to perform Authentication via PAP which is how credentials in radcheck are supposed to be handled.  Below is some of the inline documentation taken from the 'default' virtual server that Freeradius uses.

#  In general, you SHOULD NOT set the Auth-Type attribute.  The server
#  will figure it out on its own, and will do the right thing.  The
#  most common side effect of erroneously setting the Auth-Type
#  attribute is that one authentication method will work, but the
#  others will not.

I don't know that I have the skill to change how the Rest API is currently implemented but I believe it's possible to support authentication against auth_user and radcheck tables simultaneously.  I'll take a look and see if it's something I can do.

nemesisdesign: @2stacks thanks for explaining, now my understanding has increased.
Try hacking this line: https://github.com/openwisp/django-freeradius/blob/master/django_freeradius/api/views.py#L48
That's the line that explicitly sets the `Auth-Type := Reject`

2stacks: Thanks, I'll start looking there.  I'm pretty sure we will still have to set Auth-Type := Accept since the django hashes wont work with PAP.  If a user doesn't exist in the auth_user table or their credentials can't be verified I think instead of returning Auth-Type := Reject we just let it continue normal freeradius evaluation.  Not sure what not setting an Auth-Type will do but that's what I'm going to find out :)

PR #155, opened 2018-07-28
[qa] Fixed incorrect imports error

rohithasrk:


nemesisdesign:
@rohithasrk this broke the readthedocs build: https://readthedocs.org/projects/django-freeradius/builds/7562237/

PR #159, opened 2018-07-31
[tests] Remove migrations of sample app

nemesisdesign:
I don't remember if migrations for the sample app can be omitted or not.
If they can be omitted, we should do so to avoid the pain of having to maintain them.


nemesisdesign:
Doesn't work :(

PR #163, opened 2018-08-04
Making the test suite extensible to support resuability

rohithasrk:


rohithasrk:
I'm improving these tests in such a way that https://github.com/rohithasrk/openwisp-radius-prototype/tree/multitenancy-tests passes all the tests and works properly.

rohithasrk:
@nemesisdesign Need to look into the API test suite once. And then we are good to go.

rohithasrk:
@nemesisdesign It can be merged.

django_freeradius/base/models.py
```diff
@@ -710,7 +717,13 @@ def save(self):
...
-                                            attribute=attribute_map.get(attr),
-                                            value=getattr(profile, attr))
+                    options = dict(username=self.user.username,
+                                   attribute=attribute_map.get(attr),
+                                   value=getattr(profile, attr))
+                    check = self._create_radcheck(**options)
+                    check.save()
         super(AbstractRadiusUserProfile, self).save()
+
+    def _create_radcheck(self, **kwargs):
```
nemesisdesign: the method of this name is misleading because it does not create but it instantiates a new model instance.
To make it more readable you could do something like:

```python
@property
def _radcheck_model(self):
    return swapper.load_model('django_freeradius', 'RadiusCheck')
```

django_freeradius/tests/__init__.py
```diff
@@ -0,0 +1,164 @@
+class CreateObjectsMixin(object):
```
nemesisdesign: it would be better to rename this `CreateRadiusObjectsMixin` or something similar

django_freeradius/tests/utils.py
```diff
@@ -0,0 +1,14 @@
...
+
+class FileMixin(object):
+    def _get_path(self, file):
+        d = os.path.dirname(os.path.abspath(__file__))
+        return os.path.join(d, file)
+
+
+class CallCommandMixin(object):
+    def _call_command(self, command, **kwargs):
+        call_command(command, **kwargs)
```
nemesisdesign: why not move all the test mixins in the same file? Either here or in `__init__.py`? Wouldn't it make it easier to find these mixins when reading the source code?

django_freeradius/base/models.py
rohithasrk: @nemesisdesign the main reason for creating this method is to override this in openwisp-radius. i.e., add an extra organization field in the kwargs. Like https://github.com/openwisp/openwisp-radius/blob/master/openwisp_radius/models.py#L81

nemesisdesign: ok just rename it then, the name is misleading

PR #169, opened 2018-08-15
[Models] Radius Check Form Bug Fix

### LINKED ISSUE #134: [models] For fields other than the password types show the value in plain text, also while typing.
When changing the RadiusCheck model instances from the admin interface, we see that the checks enforced on the password formats come in between adding any new non password attributes.

For ex., in #110 we are using attributes like Max-Session-Limit etc as attributes that are not password types. Support needs to be added for the same.

atb00ker:
[Bug Fix] Radius Check Form
radcheck.js modified for making sure that only password field values are hidden in add radiuscheck form.
The "Max-Daily-Session", "Max-All-Session", "Max-Daily-Session-Traffic" attribute values are not hidden.
 
Fixes #134

### DIFF
--- a/django_freeradius/static/django-freeradius/js/radcheck.js
+++ b/django_freeradius/static/django-freeradius/js/radcheck.js
@@ -28,4 +28,17 @@ django.jQuery(function ($) {
         newValueRow.hide();
         rawValueRow.show();
     });
+    var value_field = $("#id_new_value");
+    function setValueInputType() {
+        var selected_attribute = $('.field-attribute option:selected').text();
+        if (selected_attribute === "Max-Daily-Session" ||
+                selected_attribute === "Max-All-Session" ||
+                selected_attribute === "Max-Daily-Session-Traffic") {
+            value_field.prop("type", "text");
+        } else {
+            value_field.prop("type", "password");
+        }
+    }
+    $("#change-radcheck-value").click(setValueInputType);
+    $(".field-attribute").change(setValueInputType);
 });


nemesisdesign:
@atb00ker next time would be great if you could include `fixes #<issue-number` in the commit message as well (commit history is easier to analyze).

atb00ker:
@rohithasrk , Can you please help me reproduce that issue? :smile:

rohithasrk:
Try clicking on the change button and start typing a value.
screenshot from 2018-08-22 21-43-36

atb00ker:
Thanks, @rohithasrk, fixed it. :smile:

django_freeradius/static/django-freeradius/js/radcheck.js
```diff
@@ -28,4 +28,21 @@ django.jQuery(function ($) {
...
+        var selected_attribute = $('.field-attribute option:selected').text();
+        if (selected_attribute === "Max-Daily-Session" ||
+                selected_attribute === "Max-All-Session" ||
+                selected_attribute === "Max-Daily-Session-Traffic") {
+            value_field.prop("type", "text");
+        } else {
+            value_field.prop("type", "password");
+        }
+    }
+    $("#change-radcheck-value").click(function () {
```
nemesisdesign: it should be possible to simplify this as:

```javascript
$("#change-radcheck-value").click(setValueInputType);
```

Can you try and let me know?

django_freeradius/static/django-freeradius/js/radcheck.js
```diff
@@ -28,4 +28,21 @@ django.jQuery(function ($) {
...
+        setValueInputType();
+    });
+    $(".field-attribute").change(function () {
```
nemesisdesign: Same here:

```javascript
$(".field-attribute").change(setValueInputType);
```

PR #172, opened 2018-09-04
[#170][limits] Make units explicit

### LINKED ISSUE #170: [limits] Make units explicit
When adding/editing a profile, the units are not explicit (eg: seconds, bytes, etc).

We should do two things:

- add `help_text` to the model fields, which will be shown in the admin
- ~~make these units explicit in the documentation page too~~ EDIT: not needed, it's already explained :+1:

atb00ker:
Added units in 'help_text'

Fixes #170


atb00ker:
Yes it did, i forgot to push them before, added!
Thanks for cehcking! :+1: :smile:

PR #174, opened 2018-09-06
[models][#167] Make username unique

### LINKED ISSUE #167: [models] Help text say "A unique ... name" but field is not unique
The `help_text` of `AbstractRadiusProfile.name` says `A unique profile name`, but the field has no `unique=True` constraint.

Same applies for `AbstractRadiusBatch`.

It would be useful to have them as unique fields, we just have to ensure that in multitenancy the constraint is removed in favour of `unique_together = ('organization', 'name')`, because different organizations may use the same name.

atb00ker:
AbstractRadiusProfile.name and
AbstractRadiusBatch.name are moved to the
concrete classes and made to be unique

Fixes #167


django_freeradius/base/models.py
```diff
@@ -513,10 +513,14 @@ def __str__(self):
 
 
 class AbstractRadiusBatch(TimeStampedEditableModel):
+    '''
+    Abstract class for making mutliple users of
+    django-radius.
+    Note: You'll (most probably) need to add the following
+    field in your concrete class:
```
nemesisdesign: simply add the unique here and get rid of this docstring, openwisp-radius will take care of doing the right thing

django_freeradius/base/models.py
```diff
@@ -646,10 +650,13 @@ def _remove_files(self):
 
 @python_2_unicode_compatible
 class AbstractRadiusProfile(TimeStampedEditableModel):
+    '''
+    Abstract class for making users of django-radius.
+    Note: You'll (most probably) need to add the following
+    field in your concrete class:
```
nemesisdesign: same here

django_freeradius/base/models.py
```diff
@@ -646,10 +650,13 @@ def _remove_files(self):
...
     name = models.CharField(verbose_name=_('name'),
-                            max_length=128,
-                            help_text=_('A unique profile name'),
-                            db_index=True)
+                        db_index=True)
```
nemesisdesign: pay attention to the indentation here

nemesisdesign:
@atb00ker just noticed now the commit message was wrong, it was not `username` but `name`

PR #193, opened 2018-10-30
[admin] Remove redundant condition check

dwang:
"and not change" is removed because the condition change is already being tested on line 299. 
This is redundant and removing it improves code quality.

Tests pass, although `./runflake8` complains about unrelated errors in `settings.py`.

### DIFF
--- a/django_freeradius/base/admin.py
+++ b/django_freeradius/base/admin.py
@@ -298,7 +298,7 @@ class AbstractRadiusBatchAdmin(TimeStampedEditableAdmin):
         strategy = data.get('strategy')
         if not change:
             if strategy == "csv":
-                if data.get('csvfile', False) and not change:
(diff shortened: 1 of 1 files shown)


dwang:
Claimed, but not submitted. I was under the impression that we should only submit after our PR is merged. Should we submit when we feel like the PR is done/approved by reviewers?

R9295:
@dwang submit as soon as you've sent the PR so we know when to check your work. Also link the task in the PR description(not the commit message) for easier access.

nemesisdesign:
@R9295 is correct, I would add an hint: only submit when the quality is good to avoid too many ping pongs and shorten the approval time but it looks to me you are already doing that so that's good.

dwang:
I see, will keep that in mind for the future. Thanks!

PR #194, opened 2018-10-30
[settings] Improve default for batch mail sender #179

### LINKED ISSUE #179: [settings] Improve default for DJANGO_FREERADIUS_BATCH_MAIL_SENDER
**Reserved for GCI**

The value of `DJANGO_FREERADIUS_BATCH_MAIL_SENDER` should default to the standard [django DEFAULT_FROM_EMAIL](https://docs.djangoproject.com/en/2.1/ref/settings/#std:setting-DEFAULT_FROM_EMAIL).

dwang:
Changes DJANGO_FREERADIUS_BATCH_MAIL_SENDER from "testing@localhost" to DEFAULT_FROM_EMAIL from the Django docs. 
DEFAULT_FROM_EMAIL is equal to "webmaster@localhost". 
Instead of setting the value directly, I chose to use this variable because it will be changed dynamically in the future.

Closes #179


dwang:
Yes, that will be fine. I will add it soon. Thanks for reviewing!

Issue #181, opened 2018-09-16
[models] Remove python_2_unicode_compatible decorators


nemesisdesign:
**Reserved for GCI**

We dropped support for python 2.7, therefore we can remove all the `python_2_unicode_compatible` occurrences.

traintraveler:
Can I take this up ?

nemesisdesign:
Hey @traintraveler, Welcome!

Aare you a Google Code-In student? If yes, please wait the official start of the program in a few hours. If not, please don't, this task is reserved for Google Code-In participants.

traintraveler:
@nemesisdesign No, I'm not a code-in student, cool.

nemesisdesign:
@traintraveler if you want to contribute you can take a look at our [Contributor's Board](https://github.com/orgs/openwisp/projects/3), thank you!

PR #196, opened 2018-10-31
[models] Remove python_2_unicode_compatible

cking100:
In python3 all strings (str) are unicode string, there is no unicode decorator needed so removed all the occurrences of of the python_2_unicode_compatible decorator, solves #181


cking100:
Sure, but should i do a separate commit removing it from docs or include it in the above one? @nemesisdesign

R9295:
@cking100 same commit.

cking100:
Done! Please have a look on it @nemesisdesign

PR #198, opened 2018-11-01
[requirements] Updated utdated versions 

### LINKED ISSUE #178: [requirements] Update xhtml2pdf version
**Task reserved for GCI**

As you can see from [requires.io](https://requires.io/github/openwisp/django-freeradius/requirements/?branch=master), the xhtml2pdf version is slightly outdated.

Let's not only update the version, but also make it flexible so it doesn't need constant updating, eg: `>=0.2.3,<0.3.0` will allow all 0.2.x versions, which should not contain backward incompatible changes.

DotSlashTX:
Updated versions of xhtml2pdf and djangoresframework #178

Closes #178

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,8 +1,8 @@
 django>=2.0,<2.2
 swapper>=1.1.0,<1.2.0
 openwisp-utils>=0.2,<0.3
-djangorestframework>=3.8.2,<3.9
+djangorestframework>=3.8.2,<3.9.0
 passlib>=1.7.1,<1.8.0
 django-filter>=2.0.0,<2.1.0
 djangorestframework-link-header-pagination>=0.1.1,<0.2.0
-xhtml2pdf==0.2.2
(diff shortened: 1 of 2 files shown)


requirements.txt
```diff
@@ -1,8 +1,8 @@
 django>=2.0,<2.2
 swapper>=1.1.0,<1.2.0
 openwisp-utils>=0.2,<0.3
-djangorestframework>=3.8.2,<3.9
+djangorestframework>=3.8.2,<3.9.0
```
nemesisdesign: the latest version of DRF is 3.9.0, hence this will not catch it. increase it to one minor version, `3.10.0`

DotSlashTX: Please check it now :) I've made the changes. Please let me know if I need to change anything else as I'm unable to submit other tasks that I've already completed. 
Thanks Sir

DotSlashTX:
Please check out now
I've squashed the commit as well

PR #200, opened 2018-11-05
[api] Improve robustness of TokenAuthentication #182

### LINKED ISSUE #182: [api] Improve robustness of TokenAuthentication
**Reserved for GCI**

At this line: https://github.com/openwisp/django-freeradius/blob/master/django_freeradius/api/views.py#L29

The code would fail if the authorization header does not contain any space. Proceed as follows:

- replicate the issue in a test case (this is also called Test Driven Development)
- once the test fails, fix the code so it doesn't fail

cking100:
Tests works fine if the authorization header is in the standard format `Authorization: <Type> <Credentials>` but would fail if there is no space in between type and credentials.

Fixes #182

### DIFF
--- a/django_freeradius/api/views.py
+++ b/django_freeradius/api/views.py
@@ -7,7 +7,7 @@ from django_filters import rest_framework as filters
 from django_filters.rest_framework import DjangoFilterBackend
 from rest_framework import generics, status
 from rest_framework.authentication import BaseAuthentication
-from rest_framework.exceptions import AuthenticationFailed, ValidationError
+from rest_framework.exceptions import AuthenticationFailed, ParseError, ValidationError
 from rest_framework.response import Response
 from rest_framework.views import APIView
 
@@ -28,7 +28,10 @@ class TokenAuthentication(BaseAuthentication):
         if request.META.get('HTTP_AUTHORIZATION', False):
             headers = request.META.get('HTTP_AUTHORIZATION').split(',')
             for header in headers:
-                token = header.split(' ')[1]
+                try:
+                    token = header.split(' ')[1]
+                except IndexError:
+                    raise ParseError('Invalid token')
                 if token == app_settings.API_TOKEN:
                     return (AnonymousUser(), None)
         if request.GET.get('token') == app_settings.API_TOKEN:
--- a/django_freeradius/tests/base/test_api.py
+++ b/django_freeradius/tests/base/test_api.py
@@ -15,6 +15,15 @@ User = get_user_model()
 
 
 class BaseTestApi(object):
+    def test_valid_token(self):
+        options = dict(username='molly', password='barbar')
+        self._create_user(**options)
+        auth_header = self.auth_header.replace(' ', '')  # removes spaces in token
+        response = self.client.post(reverse('freeradius:authorize'),
+                                    {'username': 'molly', 'password': 'barbar'},
+                                    HTTP_AUTHORIZATION=auth_header)
+        self.assertEqual(response.status_code, 400)
+
     def test_disabled_user_login(self):
         options = dict(username='barbar', password='molly', is_active=False)
         self._create_user(**options)


R9295:
@cking100 looks good!

django_freeradius/api/views.py
```diff
@@ -28,7 +28,10 @@ def authenticate(self, request):
         if request.META.get('HTTP_AUTHORIZATION', False):
             headers = request.META.get('HTTP_AUTHORIZATION').split(',')
             for header in headers:
-                token = header.split(' ')[1]
+                try:
```
AnonGuy: I wouldn't use a `try`/`except` here, there are more readable solutions to this. I'd probably do something like:
```suggestion
tokens = header.split(' ')
if len(tokens) > 1:
    token = tokens[1]
else:
    token = header[6:]
```

AnonGuy: I think we should be avoiding the usage of exceptions as flow control, as much as possible.

cking100: Line 31 would still return a IndexError in this case as there is no space i think it is necessary to catch that error so to be able to pass the test tho i might be wrong. What are your thoughts? @R9295 @nemesisdesign @AnonGuy

AnonGuy: It won't raise an IndexError, even if there isn't a space.

cking100: What are your thoughts on this? @R9295 @nemesisdesign

AnonGuy: Say `header` has the value `"foobar"`, `header.split(" ")` won't raise an IndexError, it will return the list `["foobar"]`.

AnonGuy: (There's no need to delete your messages)

cking100: Yeah you are right i won't raise any error.

nemesisdesign: @AnonGuy using try/except in python is perfectly acceptable and actually recommended unless the exceptional case is common

nemesisdesign: @cking100 I will review tomorrow

R9295:
I approved it @nemesisdesign

cking100:
yes i did @nemesisdesign

nemesisdesign:
yes but the implementation is really weird, the authentication should fail instead of trying to figure out where the token is, @cking100 I'm about to suggest an improvement to this patch

cking100:
No issues Review it and lemme know i will update it @nemesisdesign

django_freeradius/api/views.py
```diff
@@ -28,7 +28,10 @@ def authenticate(self, request):
...
+                    token = header.split(' ')[1]
+                except IndexError:
+                    token = header[6:]
```
nemesisdesign: It's not a good practice to hide an error of the client. In this case the client is sending bogus data and we need to let them know.

Do the following:

```
raise ParseError('Invalid token')
```

You will need to import `ParseError`, add it to the end of line 10.

django_freeradius/tests/base/test_api.py
```diff
@@ -15,6 +15,16 @@
...
 
 class BaseTestApi(object):
+    def test_valid_token(self):
+        options = dict(username='molly', password='barbar')
+        self._create_user(**options)
+        auth_header = self.auth_header.replace(' ', '')  # removes spaces in token
+        response = self.client.post(reverse('freeradius:authorize'),
+                                    {'username': 'molly', 'password': 'barbar'},
+                                    HTTP_AUTHORIZATION=auth_header)
+        self.assertEqual(response.status_code, 200)
```
nemesisdesign: please update your code to expect `400`

django_freeradius/tests/base/test_api.py
```diff
@@ -15,6 +15,16 @@
...
+        self.assertEqual(response.data, {'control:Auth-Type': 'Accept'})
```
nemesisdesign: remove this line

nemesisdesign:
Maybe the aim of the issue was misunderstood: the aim was not to avoid rejecting clients if a space is missing, the aim was avoid the application crashing with an internal server error. I hope this helps to clarify the situation.

django_freeradius/api/views.py
cking100: okay, so instead of fixing the invalid token it should raise a parse error of invalid token?

cking100:
Yeah, that actually makes sense as if the token is in wrong format instead of fixing every possible wrong token it will return a 400 and will also avoid app crashing

nemesisdesign:
:+1:

cking100:
Done! @nemesisdesign Please have a look when you are free thanks!

cking100:
@nemesisdesign Thank you :)

PR #203, opened 2018-11-16
[ci] Add ci check for migrations

cking100:
Checks if the migration file has a descriptive name if not, then  it exits with 1 and forces the build to fail.


.travis.yml
```diff
@@ -29,6 +29,9 @@ before_install:
   - pip install --no-cache-dir -U -r requirements-test.txt
   - ./runflake8
   - ./runisort
+  - # TODO: remove after openwisp-utils > 0.2 is released
+  - pip install https://github.com/openwisp/openwisp-utils/tarball/master#egg=openwisp_utils --upgrade
```
nemesisdesign: remove this because openwisp-utils 0.2.1 has been released

runcheckmigration
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+set -e
+checkmigrations --migrations-to-ignore 3 \
```
nemesisdesign: we don't need to ignore migrations here, remove `--migrations-to-ignore 3 \` and put the command on one line only

runcheckmigration
```diff
@@ -0,0 +1,6 @@
...
+                --migration-path ./django_freeradius/migrations/ || exit 1
+checkmigrations --migrations-to-ignore 3 \
```
nemesisdesign: same here

requirements-test.txt
```diff
@@ -11,3 +11,4 @@ freezegun
 django-extensions
 django-rest-auth>=0.9.3,<0.12.0
 django-allauth>=0.36.0,<0.39.0
+openwisp-utils
```
nemesisdesign: `openwisp-utils>=0.2.1`

cking100:
@nemesisdesign Please have a look now

PR #212, opened 2018-12-02
[accounting API] automatically fill in groupname 

### LINKED ISSUE #185: [accounting API] Automatically determine groupname
**Reserved for GCI**

The radius accounting table has a `groupname` field which cannot be sent by freeradius in the accounting query, it has to be determined in another way.

There are different way to do this, but we should use the Django ORM so this feature can be handled automatically by the API. To implement this feature we should do the following:

- when saving a new accounting, the system can figure out the groupname related to the user by selecting the related `RadiusUserGroup` with the highest priority (`1` being highest)
- a few words describing this feature should be spent in the API documentation
- this feature can be turned off with a setting if needed, the setting must be documented in the available settings page and link the related API documentation section
- a test case which saves an accounting instance and expects the `groupname` to be filled automatically should be written
- a test case which saves an accounting via the API and expects the `groupname` to be filled automatically should also be written
- a test case which tests this feature when disabled should also be written, to implement how this is done you can take a look at `django_freeradius.tests.base.test_api.BaseTestApiReject`

shashwat1002:
for accounting instance when saved from API

The groupname is of the group with the highest priority related to the user
Fixes #185

### DIFF
--- a/django_freeradius/api/views.py
+++ b/django_freeradius/api/views.py
@@ -1,3 +1,5 @@
+import logging
+
 import drf_link_header_pagination
 import swapper
 from django.contrib.auth import get_user_model
@@ -168,10 +170,11 @@ class AccountingView(generics.ListCreateAPIView):
 
     def create(self, request, *args, **kwargs):
         is_start = request.data['status_type'] == 'Start'
+        data = request.data.copy()  # import because request objects is immutable
         for field in ['session_time', 'input_octets', 'output_octets']:
             if is_start and request.data[field] == '':
-                request.data[field] = 0
-        serializer = self.get_serializer(data=request.data)
+                data[field] = 0
+        serializer = self.get_serializer(data=data)
         serializer.is_valid()
         error_keys = serializer.errors.keys()
         errors = len(error_keys)
@@ -185,6 +188,24 @@ class AccountingView(generics.ListCreateAPIView):
             return self.update(request, *args, **kwargs)
         else:
             raise ValidationError(serializer.errors)
+
+    def perform_create(self, serializer):
+        if app_settings.API_ACCOUNTING_AUTO_GROUP:
+            user_model = get_user_model()
+            username = serializer.validated_data.get('username', '')
+            try:
+                user = user_model.objects.get(username=username)
+            except User.DoesNotExist:
+                logging.info('no corresponding user found '
+                             'for username: {}'.format(username))
+                serializer.save()
+            else:
+                group = user.radiususergroup_set.order_by('priority').first()
+                # user may not have a group defined
+                groupname = group.groupname if group else None
+                serializer.save(groupname=groupname)
+        else:
+            return super().perform_create(serializer)
 
     def update(self, request, *args, **kwargs):
         try:
--- a/django_freeradius/settings.py
+++ b/django_freeradius/settings.py
@@ -32,3 +32,4 @@ API_AUTHORIZE_REJECT = getattr(settings, 'DJANGO_FREERADIUS_API_AUTHORIZE_REJECT
 SOCIAL_LOGIN_ENABLED = {'allauth.socialaccount',
                         'rest_framework.authtoken'}.issubset(settings.INSTALLED_APPS)
 DISPOSABLE_USER_TOKEN = getattr(settings, 'DJANGO_FREERADIUS_DISPOSABLE_USER_TOKEN', True)
+API_ACCOUNTING_AUTO_GROUP = getattr(settings, 'DJANGO_FREERADIUS_API_ACCOUNTING_AUTO_GROUP', True)
--- a/django_freeradius/tests/test_api.py
+++ b/django_freeradius/tests/test_api.py
@@ -7,6 +7,7 @@ from django.test import TestCase
 
 from . import CreateRadiusObjectsMixin, PostParamsMixin
 from .. import settings as app_settings
+from ..models import RadiusUserGroup
 from .base.test_api import BaseTestApi, BaseTestApiReject
 
 RadiusPostAuth = swapper.load_model('django_freeradius', 'RadiusPostAuth')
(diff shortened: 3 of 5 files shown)


django_freeradius/api/views.py
```diff
@@ -186,6 +190,20 @@ def create(self, request, *args, **kwargs):
         else:
             raise ValidationError(serializer.errors)
 
+    def perform_create(self, serializer):
+        user_model = get_user_model()
+        if app_settings.AUTOMATIC_GROUPNAME_ASSIGNMENT_RADIUSACCOUNTING:
+            try:
+                user = user_model.objects.get(username=serializer.validated_data.get('username'))
+                highest_priority_usergroup = return_highest_priority_usergroup(user)
```
nemesisdesign: this is overkill, try just the following:

```python
group = user.radiususergroup_set.order_by('priority').first()
```

get rid of the function in `utils.py`

django_freeradius/settings.py
```diff
@@ -32,3 +32,4 @@
 SOCIAL_LOGIN_ENABLED = {'allauth.socialaccount',
                         'rest_framework.authtoken'}.issubset(settings.INSTALLED_APPS)
 DISPOSABLE_USER_TOKEN = getattr(settings, 'DJANGO_FREERADIUS_DISPOSABLE_USER_TOKEN', True)
+AUTOMATIC_GROUPNAME_ASSIGNMENT_RADIUSACCOUNTING = False
```
nemesisdesign: this would make it always ``False`` regardless of configuration. the name is also too long, try the following:

```python
API_ACCOUNTING_AUTO_GROUP = getattr(settings, 'DJANGO_FREERADIUS_API_ACCOUNTING_AUTO_GROUP', True)
```
This also enables it by default, which is good because it should work in most cases and should not cause issues if enabled, but users may have their own reasons to disable it.

django_freeradius/api/views.py
```diff
@@ -168,10 +171,11 @@ def post(self, request, *args, **kwargs):
 
     def create(self, request, *args, **kwargs):
         is_start = request.data['status_type'] == 'Start'
+        data = request.data.copy()  # import because request objects is immutable
         for field in ['session_time', 'input_octets', 'output_octets']:
             if is_start and request.data[field] == '':
-                request.data[field] = 0
-        serializer = self.get_serializer(data=request.data)
+                data[field] = 0
```
nemesisdesign: why not adding the groupname here directly instead of overriding `perform_create`? Wouldn't it be simpler?

shashwat1002: I did the perform_create part, because I've read several places that it's better to do so. 
And I had more freedom as to adding try-except. If you want I can get rid of this.

django_freeradius/settings.py
shashwat1002: Just to confirm, I would put the `DJANGO_FREERADIUS_API_ACCOUNTING_AUTO_GROUP` in the main settings file and `DJANGO_FREERADIUS_API_ACCOUNTING_AUTO_GROUP` in the app's settings file?

django_freeradius/api/views.py
shashwat1002: I'll do this.

django_freeradius/api/views.py
```diff
@@ -186,6 +189,20 @@ def create(self, request, *args, **kwargs):
         else:
             raise ValidationError(serializer.errors)
 
+    def perform_create(self, serializer):
```
nemesisdesign: ok we can leave this here but you know I'm picky, please rewrite this block as follows:

```python
if app_settings.API_ACCOUNTING_AUTO_GROUP:
    def perform_create(self, serializer):
        user_model = get_user_model()
        username = serializer.validated_data.get('username', '')
        try:
            user = user_model.objects.get(username=username)
        except User.DoesNotExist:
            logging.info('no corresponding user found '
                         'for username: {}'.format(username)
            serializer.save()
        else:
            group = user.radiususergroup_set.order_by('priority').first()
            # user may not have a group defined
            groupname = group.groupname if group else None
            serializer.save(groupname=groupname)
```

django_freeradius/tests/test_api.py
```diff
@@ -24,11 +25,72 @@ class TestApi(BaseTestApi, ApiTestCase):
...
+            'session_time': '',
+            'input_octets': '',
+            'output_octets': '',
+            'nas_ip_address': '127.0.0.1',
+            'session_id': '48484',
+            'unique_id': '1515151',
+            'username': 'username1',
+        })
+        accounting_created = self.radius_accounting_model.objects.get(username='username1')
+        self.assertEquals(accounting_created.groupname, "group2")
```
nemesisdesign: use single quotes, `'group2'` for consistency, please always be consistent

docs/source/general/settings.rst
```diff
@@ -180,3 +180,13 @@ The text could be anything but should have the format string operator ``{}`` for
...
 Currently, ``DEFAULT_FROM_EMAIL`` is set to to ``webmaster@localhost``.
+
+``DJANGO_FREERADIUS_API_ACCOUNTING_AUTO_GROUP``
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+**Default**: ``True``
+
+When this setting is enabled, every accounting instance saved from the API will have its ``groupname`` attribute automatically filled in.
+The value filled in will be the ``groupname`` of the ``RadiusUserGroup`` of the highest priority among the RadiusUserGroups related to the user with the ``username`` as in the accounting instance.
+In the event there is no user in the database corresponding to the ``username`` in the accounting instance, only a logging entry will be made mentioning that no such user was found.
```
nemesisdesign: great addition, let's just replace the following: `only a logging entry will be made mentioning that no such user was found.` with ``the failure will be logged with `info` level but the accounting will be saved as usual``.

django_freeradius/tests/test_api.py
```diff
@@ -24,11 +25,72 @@ class TestApi(BaseTestApi, ApiTestCase):
...
+    @classmethod
+    def setUpClass(cls):
+        app_settings.API_ACCOUNTING_AUTO_GROUP = True
+        super(TestAutomaticGroupnameSetting, cls).setUpClass()
+
+    def test_automatic_groupname_account_enabled(self):
+        user = self.user_model.objects.create_superuser(
+            username="username1", email="admin@admin.com", password="qwertyuiop"
+        )
+        usergroup1 = self._create_radius_usergroup(groupname="group1", priority=2, username='testgroup1')
```
nemesisdesign: same here for double quotes and in the rest of the file, if single quotes are used, used single quotes everywhere

django_freeradius/api/views.py
shashwat1002: I'll do this, but I was curious why in this particular way?

shashwat1002: I tried this, but the test I wrote which is supposed to check the behavior if the  setting is disabled failed i.e. the function is filling the group name even if for the test I set the setting to False. I'll investigate further.

nemesisdesign: I see, it's because the `if` block is executed only when the app is booted, so even if you turn off the value of the setting during tests, the method will still be defined and will work anyway, I'll be more flexible and propose this other approach which should work in tests too:

```python
def perform_create(self, serializer):
    if app_settings.API_ACCOUNTING_AUTO_GROUP:
        user_model = get_user_model()
        username = serializer.validated_data.get('username', '')
        try:
            user = user_model.objects.get(username=username)
        except User.DoesNotExist:
            logging.info('no corresponding user found '
                         'for username: {}'.format(username)
            serializer.save()
        else:
            group = user.radiususergroup_set.order_by('priority').first()
            # user may not have a group defined
            groupname = group.groupname if group else None
            serializer.save(groupname=groupname)
    else:
        return super().perform_create(serializer)
```

Why this approach? basically, if the setting is turned off, we avoid altering the logic of this method entirely. It introduces less risks of unforeseen bugs.

shashwat1002: Understood @nemesisdesign , I made the changes.

shashwat1002:
I also changed the test code a little bit so that every previous test is tested with the feature on.

docs/source/general/settings.rst
```diff
@@ -180,3 +180,15 @@ The text could be anything but should have the format string operator ``{}`` for
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+**Default**: ``True``
+
+When this setting is enabled, every accounting instance saved from the API will have its ``groupname`` attribute automatically filled in.
+The value filled in will be the ``groupname`` of the ``RadiusUserGroup`` of the highest priority among the RadiusUserGroups related to the user with the ``username`` as in the accounting instance.
+In the event there is no user in the database corresponding to the ``username`` in the accounting instance, the failure will be logged with `info` level but the accounting will be saved as usual.
+
+
+
```
nemesisdesign: please don't add unnecessary blank lines, one blank line is enough, please always be consistent

docs/source/general/settings.rst
```diff
@@ -180,3 +180,12 @@ The text could be anything but should have the format string operator ``{}`` for
 It is the sender email which is also to be configured in the SMTP settings.
 The default sender email is a common setting from the `Django core settings  <https://docs.djangoproject.com/en/2.1/ref/settings/#default-from-email>`_ under ``DEFAULT_FROM_EMAIL``.
 Currently, ``DEFAULT_FROM_EMAIL`` is set to to ``webmaster@localhost``.
+
+``DJANGO_FREERADIUS_API_ACCOUNTING_AUTO_GROUP``
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
nemesisdesign: `~~~~` line is too long, it should equal the length of the text, eg:

```
``DJANGO_FREERADIUS_API_ACCOUNTING_AUTO_GROUP``
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

some ReStructuredText parsers complain when this rule is not followed.

shashwat1002: I just changed this, @nemesisdesign

PR #218, opened 2018-12-09
[admin] Remove admin NasForm in favour of customizable NAS choices #183

### LINKED ISSUE #183: [admin] Remove admin NasForm in favour of customizable nas choices
**Reserved for GCI**

The custom NAS form in the admin of django-freeradius allows to enter a NAS type which is not included in the available options, but due to how django choices work, custom values won't be shown in the admin select list.

We should do the following:

- make the NAS choices customizable with a setting
- document the new setting
- remove the current NAS form and eventual unit tests related to it

shashwat1002:
This code effectively gets rid of the  functionality in which custom NAS types could've been defined from the admin interface: instead, if an attribute called `CUSTOM_TYPES_FOR_NAS` should be defined in the settings which should have all the custom types the user wants add, and these custom types can then be accessed from the admin interface, note that these along with the pre-defined types are accessible. 
This code also change the `NasForm` and changes the testcase which expects `NasForm` to allow custom types. 

Fixes #183


docs/source/general/settings.rst
```diff
@@ -189,3 +189,11 @@ Currently, ``DEFAULT_FROM_EMAIL`` is set to to ``webmaster@localhost``.
 When this setting is enabled, every accounting instance saved from the API will have its ``groupname`` attribute automatically filled in.
 The value filled in will be the ``groupname`` of the ``RadiusUserGroup`` of the highest priority among the RadiusUserGroups related to the user with the ``username`` as in the accounting instance.
 In the event there is no user in the database corresponding to the ``username`` in the accounting instance, the failure will be logged with `info` level but the accounting will be saved as usual.
+
+``CUSTOM_TYPES_FOR_NAS``
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+**Please note: this settings won't exist by default**
```
nemesisdesign: change this to:

```
**Default**: `tuple()`
```

docs/source/general/settings.rst
```diff
@@ -189,3 +189,11 @@ Currently, ``DEFAULT_FROM_EMAIL`` is set to to ``webmaster@localhost``.
...
+
+This setting should have a tuple of tuples, where each tuple should be an option for `ChoiceField` for the NAS form. This setting should be used to allow custom NAS types to be selected from the admin interface
```
nemesisdesign: should be:

```
`ChoiceField`
```

Try also to keep the lines shorter please.

django_freeradius/settings.py
```diff
@@ -33,3 +33,4 @@
                         'rest_framework.authtoken'}.issubset(settings.INSTALLED_APPS)
 DISPOSABLE_USER_TOKEN = getattr(settings, 'DJANGO_FREERADIUS_DISPOSABLE_USER_TOKEN', True)
 API_ACCOUNTING_AUTO_GROUP = getattr(settings, 'DJANGO_FREERADIUS_API_ACCOUNTING_AUTO_GROUP', True)
+CUSTOM_TYPES_EXTRA = getattr(settings, 'CUSTOM_TYPES_FOR_NAS', None)
```
nemesisdesign: change this to:

```python
EXTRA_NAS_TYPES = getattr(settings, 'DJANGO_FREERADIUS_EXTRA_NAS_TYPES', tuple())
```

Adapt the naming in the other files in order to be consistent with this suggesiton.
Note: default to empty tuple so we can avoid the `if` in `models.py`.

docs/source/general/settings.rst
```diff
@@ -189,3 +189,11 @@ Currently, ``DEFAULT_FROM_EMAIL`` is set to to ``webmaster@localhost``.
...
+when creating a new NAS instance.
```
nemesisdesign: If you could supply an example for our users it would be great, consider that many of our users are not familiar with django and some are not even familiar with python.

django_freeradius/base/models.py
```diff
@@ -90,6 +90,12 @@
 )
 
 
+if app_settings.CUSTOM_TYPES_EXTRA:  # if cutom NAS types have been defined.
+    RAD_NAS_TYPES_ALL = RAD_NAS_TYPES + app_settings.CUSTOM_TYPES_EXTRA
+else:
+    RAD_NAS_TYPES_ALL = RAD_NAS_TYPES
```
nemesisdesign: now after you deafult the new setting to an empty tuple you can avoid the `if`, as a consequence the coverage should not decrease.

docs/source/general/settings.rst
nemesisdesign: PS: remember to build the docs and look at it in your browser before committing, it will help you to spot if there's any mistake or anything else to improve

shashwat1002:
Thanks! @nemesisdesign , just made the changes 😄

PR #221, opened 2018-12-10
 [admin] Batch User creation form admin fixed  #175

### LINKED ISSUE #175: [bug] Batch User creation form throwing `TypeError`
When the form is submitted without entering `Number of users` ; `TypeError at /admin/django_freeradius/radiusbatch/add/` is received due to the fact that `NoneType` object cannot be interpreted as an integer in the `prefix_generate_user()`, a check for the same is required.

shashwat1002:
A test has also been added corresponding to the issue. 
Fixes #175


django_freeradius/base/forms.py
```diff
@@ -59,6 +59,15 @@ class RadiusBatchForm(forms.ModelForm):
                                          validators=[MinValueValidator(1)],
                                          help_text=_('Number of users to be generated'))
 
+    def clean(self):
+        data = self.cleaned_data
+        strategy = data.get('strategy')
+        if strategy == 'prefix':
```
nemesisdesign: can you simplify this please?

```python
if data.get('strategy') == 'prefix' and not data.get('number_of_users'):
    # ...
```

django_freeradius/api/serializers.py
```diff
@@ -90,7 +90,16 @@ class RadiusBatchSerializer(serializers.ModelSerializer):
...
+                                               min_value=1,)
+
+    def validate(self, data):
+
+        if data['strategy'] == 'prefix' and data.get('number_of_users', '') == '':
+            raise serializers.ValidationError(
+                {'number_of_users': 'The field number_of_users cannot be empty'}
+            )
+        else:
+            return super(RadiusBatchSerializer, self).validate(data)
```
nemesisdesign: can you try replacing this with:

```python
def validate(self, data):
    if data['strategy'] == 'prefix' and not data.get('number_of_users'):
        raise serializers.ValidationError({
            'number_of_users': 'The field number_of_users cannot be empty'
        })
    return super().validate(data)
```

we can use the form `super()` because this module is python3 only.

django_freeradius/api/serializers.py
```diff
@@ -90,7 +90,16 @@ class RadiusBatchSerializer(serializers.ModelSerializer):
     pdf = serializers.FileField(required=False, read_only=True)
     number_of_users = serializers.IntegerField(required=False,
                                                write_only=True,
-                                               min_value=1)
+                                               min_value=1,)
```
nemesisdesign: why did you add this comma? Can you remove it please?

django_freeradius/api/serializers.py
shashwat1002: I think `data.get('number_of_users')` should be replaced with `data.get('number_of_users`, '')` because I noticed that if the api request dosen't have the `number_of_users` attribute passed in then it's not available as a key in `data` @nemesisdesign .

django_freeradius/api/serializers.py
shashwat1002: My bad, @nemesisdesign . I'll remove it just now.

django_freeradius/api/serializers.py
nemesisdesign: then it would return `None` which would be falsy so that should be ok

shashwat1002:
Done @nemesisdesign

Issue #171, opened 2018-08-26
[docs] Add management commands suggestion in installation page


nemesisdesign:
**Reserved for GCI**

Some management commands are essential to make some features work (eg: deactivate and delete expired users).

We should recommend to add all the management commands in the crontab, they are all needed in a real world deployment. This should be done in the [setup page](https://django-freeradius.readthedocs.io/en/latest/general/setup.html).

atb00ker:
The management commands seem to have their own [page](https://django-freeradius.readthedocs.io/en/latest/general/management_commands.html).

Shoud i add:
```
Note: Some management commands are necessary for enabling certain features and for database cleanup.
It recommended to have these commands in crontab in production environments.
[Find out more about these commands in its dedicated page](https://django-freeradius.readthedocs.io/en/latest/general/management_commands.html)
```
or should i add them in the same page separately.

nemesisdesign:
@atb00ker yes that would be good, where would you add that text?

PR #222, opened 2018-12-11
[docs] Add management commands suggestion in the setup page #171

dwang:
A suggestion is made in the setup page to automate the usage of the management commands. A sample crontab is also added.

### DIFF
--- a/docs/source/general/setup.rst
+++ b/docs/source/general/setup.rst
@@ -169,3 +169,56 @@ Run tests with:
 .. code-block:: shell
 
     ./runtests.py
+
+Automating management commands
+------------------------------
+
+Some management commands are necessary to enable certain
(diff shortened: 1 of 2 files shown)


docs/source/general/setup.rst
```diff
@@ -169,3 +169,56 @@ Run tests with:
...
+installed for development, navigate to the base directory of 
+the cloned fork. Then, run:
+
+.. code-block:: shell
+    
+    cd tests/
+    pwd
+
+More information can be found at
+the `management commands page <https://django-freeradius.readthedocs.io/en/latest/general/management_commands.html>`_.
```
nemesisdesign: can you change this so it's a relative path instead of an absolute URL please?

`management commands page <./management_commands.html>`_

PR #225, opened 2018-12-13
[docs] Specifies user to keep the django server running #223

### LINKED ISSUE #223: [docs] Clarify django application must be running before starting freeradius
In the freeradius installation and configuration doc page it is not stated explicitly that the django application must be running before freeradius is started. This implicit assumption is throwing new users off track as shown by this post: https://groups.google.com/d/msg/openwisp/UvK23cJW_3k/1DCvk0EKDgAJ

We should add one or more notes regarding this, probably a note in the `debugging` section would also help.

YashJipkate:
The docs currently assume that the user is running the django-freeradius server in the background. Since the docs did not have any instances of pointing that, this commit will make the doc mention clearly the need to keep the django-freeradius server running in the background while attempting to start the freeradius server with the config changes specified in the docs. Fixes #223

### DIFF
--- a/docs/source/general/freeradius.rst
+++ b/docs/source/general/freeradius.rst
@@ -283,6 +283,11 @@ Restart freeradius to load the new configuration:
     # alternatively if you are using systemd
     systemctl restart freeradius
 
+In case of any error, run ``journalctl -xe`` to know the cause. You can also run ``freeradius -X`` to know the 
+detailed cause. One error might be that the django-freeradius server might not be running in the background.
+To set up the django-freeradius server refer `this <https://django-freeradius.readthedocs.io/en/latest/general/setup.html#installing-for-development>`_
+Also make sure that this server runs on the port specified in ``/etc/freeradius/mods-enabled/rest``.
+
 You may also want to take a look at the `Freeradius documentation <http://freeradius.org/doc/>`_
 for further details on how to configure other modules.
 


docs/source/general/freeradius.rst
```diff
@@ -283,6 +283,11 @@ Restart freeradius to load the new configuration:
     # alternatively if you are using systemd
     systemctl restart freeradius
 
+In case of any error, run ``journalctl -xe`` to know the cause. You can also run ``freeradius -X`` to know the 
+detailed cause. One error might be that the django-freeradius server might not be running in the background.
+To set up the django-freeradius server refer `this <https://github.com/YashJipkate/django-freeradius.git>`_
```
ppabcd: why you referer django-freeradius to your repo?

YashJipkate: Oh I am sorry, didn't notice that blunder. Seems like its a bad day.
Thanks for pointing that out @ppabcd !

YashJipkate: I have corrected it and pushed again. Please review it.
Thanks.

.gitignore
```diff
@@ -69,3 +69,6 @@ media/
 
 #Pycharm projects
 .idea/
+
+#VSCode Sphinx Extension
+docs/.vscode/settings.json
```
ppabcd: I think it does not need to be included. Not everyone uses vscode. Keep this settings in your local .gitignore without push to github repository.

YashJipkate: Ok thanks for pointing that out!
I have made the necessary changes and pushed again. Please review it.
Thanks

PR #235, opened 2019-05-22
[django] Improve support for django 2.2 #231

### LINKED ISSUE #231: [django] Add support for django 2.2
Add support for django 2.2.

Vivekrajput20:
closes #231

### DIFF
--- a/django_freeradius/base/admin.py
+++ b/django_freeradius/base/admin.py
@@ -1,9 +1,10 @@
 import swapper
 from django.contrib import messages
 from django.contrib.admin import ModelAdmin, StackedInline
-from django.contrib.admin.actions import delete_selected
 from django.contrib.admin.templatetags.admin_static import static
+from django.contrib.admin.utils import model_ngettext
 from django.contrib.auth.admin import UserAdmin as BaseUserAdmin
+from django.core.exceptions import PermissionDenied
 from django.utils.translation import ugettext_lazy as _
 from openwisp_utils.admin import ReadOnlyAdmin, TimeReadonlyAdminMixin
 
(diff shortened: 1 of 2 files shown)


Vivekrajput20:
@nemesisdesign  Please revert the changes made in the previous PR and review this one.

django_freeradius/base/admin.py
```diff
@@ -141,16 +142,32 @@ def has_delete_permission(self, request, obj=None):
             return False
         return super().has_delete_permission(request, obj)
 
-    def delete_selected_batch(self, request, queryset):
+    def delete_selected_groups(self, request, queryset):
```
nemesisdesign: See here: https://stackoverflow.com/questions/10814972/override-delete-selected-and-update-fields-of-child-model

Maybe we can simply name it `delete_selected`. What do you think?

django_freeradius/base/admin.py
```diff
@@ -141,16 +142,32 @@ def has_delete_permission(self, request, obj=None):
...
+        if n:
+            queryset.delete()
+            self.message_user(request, _("Successfully deleted %(count)d %(items)s.") % {
+                "count": n, "items": model_ngettext(self.opts, n)
+            }, messages.SUCCESS)
+        return None
+
+    delete_selected_groups.allowed_permissions = ('delete',)
+
+    def get_actions(self, request):
```
nemesisdesign: if my previous suggestion is correct, we may be able to remove this

django_freeradius/base/admin.py
```diff
@@ -273,6 +290,12 @@ def delete_model(self, request, obj):
         obj.users.all().delete()
         super(AbstractRadiusBatchAdmin, self).delete_model(request, obj)
 
+    def get_actions(self, request):
```
nemesisdesign: same here

django_freeradius/base/admin.py
Vivekrajput20: I had tried to overwrite it but with the introduction of [admin.E130](https://docs.djangoproject.com/en/2.2/ref/checks/#modeladmin) we can't name any other action 'delete selected' and I had no choice but to rename the custom action.

Vivekrajput20:
@nemesisdesign I tried calling Django's default 'delete_selected' inside 'delete_selected_groups' but Idk why it didn't work. So I had to mostly rewrite it in  'delete_selected_groups'.

I would suggest review it and test it asap as the last PR that you merged is buggy. It should be reverted back or changed asap.

Vivek

PR #237, opened 2019-05-24
[authentication] Add RadiusToken #232

### LINKED ISSUE #232: [authentication] Add RadiusToken
The idea of creating a temporary token to use as a password in captive portals is very handy, but now we are using the rest framework Token model to perform the authorization in freeradius, but in order to completely separate the authentication of a user on the frontend apps that consume the openwisp API and the authorization of users in the wifi service, we should introduce a new `RadiusToken` model that will be dedicated to this purpose, tokens will be disposable: they are generated during authentication of the user and once they are used (in the freeradius authorize view) they are thrown away.

Vivekrajput20:
closes #232


django_freeradius/base/models.py
```diff
@@ -819,3 +820,36 @@ def _remove_files(self):
...
+                             on_delete=models.CASCADE,
+                             blank=True,
+                             null=True)
+
+    created = models.DateTimeField(_("Created"), auto_now_add=True)
+
+    class Meta:
+        db_table = 'radtoken'
+        verbose_name = _("radius token")
+        verbose_name_plural = _("radius tokens")
```
nemesisdesign: aren't we using single quotes in the file? Please keep the style of quotes consistent

django_freeradius/admin.py
```diff
@@ -66,6 +67,11 @@ class RadiusBatchAdmin(AbstractRadiusBatchAdmin):
     pass
 
 
+@admin.register(RadiusToken)
```
nemesisdesign: I think this can be shown only when `DEBUG` is `True`, because normally administrators won't have to deal with this

django_freeradius/base/models.py
```diff
@@ -819,3 +820,36 @@ def _remove_files(self):
         path = getattr(self, strategy_filemap.get(self.strategy)).path
         if os.path.isfile(path):
             os.remove(path)
+
+
+class AbstractRadiusToken(AutoUsernameMixin, models.Model):
+    key = models.CharField(_("Key"), max_length=40, primary_key=True)
+
+    username = models.CharField(verbose_name=_('username'),
```
nemesisdesign: why do we need this?
this looks like it shouldn't be here. We only need the key.

django_freeradius/base/models.py
```diff
@@ -819,3 +820,36 @@ def _remove_files(self):
         path = getattr(self, strategy_filemap.get(self.strategy)).path
         if os.path.isfile(path):
             os.remove(path)
+
+
+class AbstractRadiusToken(AutoUsernameMixin, models.Model):
```
nemesisdesign: Use this as a base: https://github.com/openwisp/openwisp-utils/blob/master/openwisp_utils/base.py#L8

and get rid of `created` here, it will be inherited automatically

django_freeradius/base/models.py
```diff
@@ -819,3 +820,36 @@ def _remove_files(self):
...
+    key = models.CharField(_("Key"), max_length=40, primary_key=True)
+
+    username = models.CharField(verbose_name=_('username'),
+                                max_length=64,
+                                db_index=True,
+                                blank=True)
+
+    user = models.ForeignKey(settings.AUTH_USER_MODEL,
```
nemesisdesign: I think we should opt for a `OneToOneField`, so we make sure that each user can have max 1 token.
Instead of creating a new one, we'll have to update the existing one.
Otherwise the DB will get full of old tokens and we should implement a clean up operation, but we can avoid that by making sure we will always have max one token per user.

django_freeradius/base/models.py
nemesisdesign: PS keep the quote style consistent

django_freeradius/base/models.py
Vivekrajput20: Yes, I just realized it should be one to one field. I'll fix it.

django_freeradius/base/models.py
Vivekrajput20: @nemesisdesign  do we need both the username and user field in this model like other models, or should I just drop the username field?

Vivekrajput20:
@nemesisdesign  I have updated the PR. Please review it.
I'll update the authorisation next and I think radius user token will be fine for the documentation

Vivek

django_freeradius/base/models.py
nemesisdesign: drop it, is not needed

django_freeradius/base/models.py
```diff
@@ -819,3 +820,39 @@ def _remove_files(self):
...
             os.remove(path)
+
+
+class AbstractRadiusToken(AutoUsernameMixin, TimeStampedEditableModel, models.Model):
+    id = None
+
+    modified = None
+
+    key = models.CharField(_('Key'), max_length=40, primary_key=True)
+
```
nemesisdesign: don't leave blank lines between field declaration please

django_freeradius/static/django-freeradius/css/radiustoken.css
```diff
@@ -0,0 +1,3 @@
+.form-row.field-username {
+    display: none
+}
```
nemesisdesign: this is not needed, please remove this CSS and its definition

Vivekrajput20:
@nemesisdesign should I also update the [obtain-auth-token](https://github.com/openwisp/django-freeradius/blob/master/django_freeradius/api/views.py#L273-L309) and remove REST_USER_TOKEN_ENABLED from the settings  in the same PR?

Vivekrajput20:
@nemesisdesign  can you please confirm so that I can proceed.

Vivekrajput20:
@nemesisdesign  please review the changes

django_freeradius/base/models.py
```diff
@@ -819,3 +820,29 @@ def _remove_files(self):
...
+        verbose_name_plural = _('radius token')
+        abstract = True
+
+    def save(self, *args, **kwargs):
+        if not self.key:
+            self.key = self.generate_key()
+        return super().save(*args, **kwargs)
+
+    def generate_key(self):
+        return binascii.hexlify(os.urandom(20)).decode()
```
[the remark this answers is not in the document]
Vivekrajput20: @nemesisdesign  yes I also think it'll be better to use get_random_string. I'll make the changes.

Vivekrajput20:
yes it's ready @nemesisdesign

django_freeradius/base/models.py
```diff
@@ -819,3 +819,29 @@ def _remove_files(self):
         path = getattr(self, strategy_filemap.get(self.strategy)).path
         if os.path.isfile(path):
             os.remove(path)
+
+
+class AbstractRadiusToken(TimeStampedEditableModel, models.Model):
+    id = None
+    modified = None
```
nemesisdesign: why are you removing the `id` and `modified` fields? Is there a specific reason?

Maybe in this case we can inherit from [UUIDModel](https://github.com/openwisp/openwisp-utils/blob/master/openwisp_utils/base.py#L21):

```python
from openwisp_utils.base import UUIDModel


class AbstractRadiusToken(UUIDModel):
    # ...
```

Vivekrajput20: Because the key is itself a primary key and `id` will just be a redundant field. And I think modified field in the token model will hardly be of any use as the tokens are created and then deleted on authentication and admins will not be modifying the tokens manually, they will simply regenerate them if required.
@nemesisdesign

Vivekrajput20:
@nemesisdesign  I rebased it.

django_freeradius/base/models.py
nemesisdesign: ok, just add a comment in the code regarding this detail please

tests/sample_radius/migrations/0003_radius_token.py
```diff
@@ -0,0 +1,35 @@
...
...ss - IEEE 802.11'), ('Token-Ring', 'Token-Ring'), ('FDDI', 'FDDI'), ('Wireless - CDMA2000', 'Wireless - CDMA2000'), ('Wireless - UMTS', 'Wireless - UMTS'), ('Wireless - 1X-EV', 'Wireless - 1X-EV'), ('IAPP', 'IAPP'), ('FTTP', 'FTTP'), ('IEEE 802.16', 'Wireless - IEEE 802.16'), ('IEEE 802.20', 'Wireless - IEEE 802.20'), ('IEEE 802.22', 'Wireless - IEEE 802.22'), ('PPPoA', 'PPPoA - PPP over ATM'), ('PPPoEoA', 'PPPoEoA - PPP over Ethernet over ATM'), ('PPPoEoE', 'PPPoEoE - PPP over Ethernet over Ethernet'), ('PPPoEoVLAN', 'PPPoEoVLAN - PPP over Ethernet over VLAN'), ('PPPoEoQinQ', 'PPPoEoQinQ - PPP over Ethernet over IEEE 802.1QinQ'), ('xPON', 'xPON - Passive Optical Network'), ('Wireless - XGP', 'Wireless - XGP'), ('WiMAX', ' WiMAX Pre-Release 8 IWK Function'), ('WIMAX-WIFI-IWK', 'WIMAX-WIFI-IWK: WiMAX WIFI Interworking'), ('WIMAX-SFF', 'WIMAX-SFF: Signaling Forwarding Function for LTE/3GPP2'), ('WIMAX-HA-LMA', 'WIMAX-HA-LMA: WiMAX HA and or LMA function'), ('WIMAX-DHCP', 'WIMAX-DHCP: WIMAX DCHP service'), ('WIMAX-LBS', 'WIMAX-LBS: WiMAX location based service'), ('WIMAX-WVS', 'WIMAX-WVS: WiMAX voice service'), ('Other', 'Other')], default='other', max_length=30, verbose_name='type'),
```
nemesisdesign: this is weird, why is this data migration here?
If this change is necessary at all, I would prefer if it's performed separately. What do you think?

Vivekrajput20: Yes, that should be performed separately. I think someone didn't migrate the sample_radius models, last time he made changes to the models, that's why it is there. I'll separate both the changes.

Vivekrajput20:
@nemesisdesign  I made the requested changes.

PR #240, opened 2019-07-10
[bug] Fix failing tests in openwisp_radius #236

### LINKED ISSUE #236: [bug] IntegrityError at Radius batch add operation
Radius batch add operation gives 500 Internal server error when we try to add operation with a name that already exists.

Vivekrajput20:
closes #236

### DIFF
--- a/django_freeradius/base/forms.py
+++ b/django_freeradius/base/forms.py
@@ -65,6 +65,7 @@ class RadiusBatchForm(forms.ModelForm):
         number_of_users = data.get('number_of_users')
         if strategy == 'prefix' and not number_of_users:
             self.add_error('number_of_users', 'This field is required')
+        super().clean()
         return data
 
     def __init__(self, *args, **kwargs):
--- a/django_freeradius/base/models.py
+++ b/django_freeradius/base/models.py
@@ -26,6 +26,8 @@ from django_freeradius.utils import (
 
 from .. import settings as app_settings
 
+User = get_user_model()
+
 RADOP_CHECK_TYPES = (('=', '='),
                      (':=', ':='),
                      ('==', '=='),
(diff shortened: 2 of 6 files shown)


Vivekrajput20:
@nemesisdesign  most of the tests were failing because the CSV file used in the tests had multiple entries with the same email and openwisp users has unique=True for email. I changed the email to fix that and will open another PR in openwisp radius to change Batch Add operation to add organization to the email if the duplicate email is found in CSV file.

django_freeradius/base/models.py
```diff
@@ -796,6 +777,29 @@ def prefix_add(self, prefix, n, password_length=BATCH_DEFAULT_PASSWORD_LENGTH):
         self.full_clean()
         self.save()
 
+    def get_or_create_user(self, row, users_list, generated_passwords, password_length):
```
Vivekrajput20: seprated this logic from add function  to get or create user conditionally. This will help extended app like openwisp_radius to skip creating user in case of email exists and email field is unique.

django_freeradius/base/forms.py
```diff
@@ -65,6 +65,7 @@ def clean(self):
         number_of_users = data.get('number_of_users')
         if strategy == 'prefix' and not number_of_users:
             self.add_error('number_of_users', 'This field is required')
+        super().clean()
```
Vivekrajput20: To call the unique field validators when accessing from admin site

django_freeradius/base/models.py
```diff
@@ -796,6 +777,29 @@ def prefix_add(self, prefix, n, password_length=BATCH_DEFAULT_PASSWORD_LENGTH):
...
+        User = get_user_model()
```
nemesisdesign: can we move this at the top of the file to avoid it redefine it each time? The `User` model is always only 1.

django_freeradius/base/models.py
nemesisdesign: :+1:

Vivekrajput20:
@nemesisdesign  I made the requested changes, also I slightly changed the get_or_create_user function to skip passing the generated_passwords list as param and to return user and generated_password. Now the password is appended to the list in the add function only.
Vivek

PR #241, opened 2019-07-19
[api] Add API endpoint to validate user token #238

### LINKED ISSUE #238: [api] Add API endpoint to validate user token
To implement the auto login feature in the frontend app. We'll be required to validate token stored in the signed cookie. This API endpoint will validate the user token.

Vivekrajput20:
Closes #238

### DIFF
--- a/django_freeradius/api/views.py
+++ b/django_freeradius/api/views.py
@@ -278,7 +278,15 @@ if app_settings.REST_USER_TOKEN_ENABLED:
             model = Token
             fields = ('key',)
 
-    class ObtainAuthTokenView(BaseObtainAuthToken):
+    class RadiusTokenMixin(object):
+        def get_or_create_radius_token(self, user):
+            """
+            Designed to be overridden by extensions
+            """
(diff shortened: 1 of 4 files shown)


django_freeradius/api/views.py
```diff
@@ -316,3 +316,32 @@ def post(self, request, *args, **kwargs):
...
+
+    class ValidateAuthTokenView(generics.CreateAPIView):
+        radius_token = RadiusToken
+
+        def get_or_create_radius_token(self, user):
+            """
+            Designed to be overridden by extensions
+            """
+            radius_token, rad_token_created = self.radius_token.objects.get_or_create(user=user)
+            return radius_token
```
nemesisdesign: please avoid repeating the code of this method which is copied from `ObtainAuthTokenView` by creating a mixin which is added to both views

Vivekrajput20:
@nemesisdesign I have made the requested changes.

nemesisdesign:
@Vivekrajput20 I think some work may be needed also in openwisp-radius to support this, or not?

PR #245, opened 2019-11-07
#sudo install

singh728om:
We can make true for sudo installation mode.


IAmAMissingColon:
Is This Change Required ?

atb00ker:
@singh728om If this is WIP for something or I missed some reason behind the change, please feel free to open it again / explain the change because right now, I don't understand the requirment for this change.

PR #249, opened 2019-11-12
[gitignore] Updated gitignore file to proposed standard

TheOneAboveAllTitan:
Added vitualenv specific directories to gitignore along with formatting gitignore in new propsed standard.

New standard for gitgnore for Python/Django projects proposes to have ignores in following order :

| Categories | Description |
| ------------- | ------------- |
| Project specifics   | Files/directories which are specific to a project.  |
| Django ignores  | Conventional Django ignores  |
| Python ignores | Conventional Python ignores |
| Virtual environment ignores | Possible name of directories used for creating virtual environment |
| Other | OS related and other ignores |
| Editor ignores | Directories containing IDE or editor configurations.  |

`Conventional ignores` refers to ignores included in `.gitignore` generated by [gitignore.io](gitignore.io) which I have assumed to be standard in their domains.

If approved, this format will be used for other Python project repositories of OpenWisp also.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,36 @@
+### Project Specifics ###
+tests/.env
+
+### other ###
+*.DS_Store*
+*~
+._*
+*.tar.gz
+
+### editors ###
+*.komodoproject
+.vscode/
+.idea/
+
+# Created by https://www.gitignore.io/api/venv,django
+# Edit at https://www.gitignore.io/?templates=venv,django
+
+### Django ###
+*.log
+*.pot
+*.pyc
+__pycache__/
+local_settings.py
+db.sqlite3
+db.sqlite3-journal
+*.db
+media/
+
+### Python ###
 # Byte-compiled / optimized / DLL files
 __pycache__/
 *.py[cod]
+*$py.class
 *.swp
 
 # C extensions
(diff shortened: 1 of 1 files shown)


okraits:
I'm not sure about this. It adds a lot of overhead for things a lot of people don't have.

TheOneAboveAllTitan:
> Also leave this auto generated part unchanged and add all custom sections before or after. 

There are some entries in existing file which are not present in auto-generated part, though they belong to Python portion of the file, due to this it was needed to edit that.

.gitignore
```diff
@@ -36,36 +54,92 @@ pip-delete-this-directory.txt
...
+.env
+.venv
+env/
+venv/
+ENV/
+env.bak/
+venv.bak/
+pip-selfcheck.json
+
+### other ###
```
TheOneAboveAllTitan: Are these necessary? I just added these because they were present in present file.

TheOneAboveAllTitan:
Changes suggested by @2stacks have been made. Now custom part resides at top followed by slightly modified auto-generated portion for Django and venv created using gitignore.io . Comments to provide attribution to gitignore has also been added.

2stacks:
>There are some entries in existing file which are not present in auto-generated part

I don't see any that relate to python/django when using the automated file.

That said, though I agree there's value in standardizing, it may not be worth the effort, to add a lot of things that don't obviously benefit the project, if it takes making a lot of edits to an auto generated file.

nemesisdesign:
It seems to me there's not a strong sense of need of this change right now, hence I'm closing it.
If anyone thinks implementing this kind of change may bring specific benefits to OpenWISP, please explain and we'll be happy to reconsider.

Thanks @TheOneAboveAllTitan for taking the time to contribute.

PR #256, opened 2019-12-14
Add freeradius-server-3.0.19 db fields #244

### LINKED ISSUE #244: Add freeradius-server-3.0.19 db fields
freeradius-server-3.0.19 db add feild framedinterfaceid, framedipv6address and so on now.

Is this project still devloping ?  will django model add feild soon?

Weirdo914:
Added new fields from Freeradius server (3.0.20)
Closes #244

### DIFF
--- a/django_freeradius/base/models.py
+++ b/django_freeradius/base/models.py
@@ -25,6 +25,7 @@ from django_freeradius.utils import (
 )
 
 from .. import settings as app_settings
+from .validators import ipv6_network_validator
 
 User = get_user_model()
 
@@ -398,12 +399,33 @@ class AbstractRadiusAccounting(models.Model):
                                        blank=True)
     framed_ip_address = models.GenericIPAddressField(verbose_name=_('framed IP address'),
                                                      db_column='framedipaddress',
-                                                     db_index=True,
                                                      # the default MySQL freeradius schema defines
                                                      # this as NOT NULL but defaulting to empty string
                                                      # but that wouldn't work on PostgreSQL
                                                      null=True,
                                                      blank=True)
+    framed_ipv6_address = models.GenericIPAddressField(verbose_name=_('framed IPv6 address'),
+                                                       db_column='framedipv6address',
+                                                       protocol='IPv6',
+                                                       null=True,
+                                                       blank=True)
+    framed_ipv6_prefix = models.CharField(verbose_name=_('framed IPv6 prefix'),
+                                          max_length=44,
+                                          db_column='framedipv6prefix',
+                                          validators=[ipv6_network_validator],
+                                          null=True,
+                                          blank=True)
+    framed_interface_id = models.CharField(verbose_name=_('framed interface ID'),
+                                           max_length=19,
+                                           db_column='framedinterfaceid',
+                                           null=True,
+                                           blank=True)
+    delegated_ipv6_prefix = models.CharField(verbose_name=_('delegated IPv6 prefix'),
+                                             max_length=44,
+                                             db_column='delegatedipv6prefix',
+                                             validators=[ipv6_network_validator],
+                                             null=True,
+                                             blank=True)
 
     def save(self, *args, **kwargs):
         if not self.start_time:
--- a/django_freeradius/base/validators.py
+++ b/django_freeradius/base/validators.py
@@ -1,0 +1,12 @@
+from ipaddress import IPv6Network, ip_network
+
+from django.core.exceptions import ValidationError
+
+
+def ipv6_network_validator(value):
+    try:
+        network = ip_network(value)
+    except Exception as e:
+        raise ValidationError('Invalid ipv6 prefix: {}'.format(e))
+    if not isinstance(network, IPv6Network):
+        raise ValidationError('{} is not an IPv6 prefix'.format(value))
--- a/django_freeradius/tests/base/test_api.py
+++ b/django_freeradius/tests/base/test_api.py
@@ -154,7 +154,11 @@ class BaseTestApi(object):
         'terminate_cause': 'User_Request',
         'service_type': 'Login-User',
         'framed_protocol': 'test',
-        'framed_ip_address': '127.0.0.1'
+        'framed_ip_address': '127.0.0.1',
+        'framed_ipv6_address': '::1',
+        'framed_ipv6_prefix': '0::/64',
+        'framed_interface_id': '0000:0000:0000:0001',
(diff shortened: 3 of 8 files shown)


Weirdo914:
Why is that error coming?

2stacks:
Looks like something in sample app tests.

django.db.utils.OperationalError: (1054, "Unknown column 'sample_radius_radiusaccounting.framedipv6address' in 'field list'")

Weirdo914:
> Looks like something in sample app tests.
> 
> django.db.utils.OperationalError: (1054, "Unknown column 'sample_radius_radiusaccounting.framedipv6address' in 'field list'")

I saw that but I can't understand why is it coming?

django_freeradius/base/models.py
```diff
@@ -399,11 +399,26 @@ class AbstractRadiusAccounting(models.Model):
     framed_ip_address = models.GenericIPAddressField(verbose_name=_('framed IP address'),
                                                      db_column='framedipaddress',
                                                      db_index=True,
-                                                     # the default MySQL freeradius schema defines
-                                                     # this as NOT NULL but defaulting to empty string
-                                                     # but that wouldn't work on PostgreSQL
```
2stacks: Did you mean to remove this comment?

Weirdo914: yes because it no longer as not null in freeradius schema

Weirdo914: https://github.com/FreeRADIUS/freeradius-server/blob/master/raddb/mods-config/sql/main/postgresql/schema.sql#L41

nemesisdesign: @Weirdo914, it's still NOT NULL in the mysql schema: https://github.com/FreeRADIUS/freeradius-server/blob/master/raddb/mods-config/sql/main/mysql/schema.sql#L42, although it defaults to empty string but the information in the comment is still useful to remember why it's done this way. Please put back this comment.

django_freeradius/base/models.py
```diff
@@ -399,11 +399,26 @@ class AbstractRadiusAccounting(models.Model):
...
                                                      null=True,
                                                      blank=True)
+    framed_ipv6_address = models.GenericIPAddressField(verbose_name=_('framed IPv6 address'),
+                                                       db_column='framedipv6address',
+                                                       db_index=True,
```
nemesisdesign: let's remove `db_index=True` from `framed_ipv6_address` but also `framed_address` since it's not in the original schema and the index is not being used right now by any query in the application.

django_freeradius/tests/base/test_api.py
```diff
@@ -181,7 +185,7 @@ def assertAcctData(self, ra, data):
         to ensure they match
         """
         for key, value in data.items():
-            if key == 'status_type':
+            if key in ('status_type', 'framed_ipv6_address'):
```
Weirdo914: Because of the way django treats ipv6 addresses, it keeps converting `::ffff:c0a8:b603` into `::ffff:192.168.182.3`, so I could only do this.

django_freeradius/base/models.py
Weirdo914: Done

django_freeradius/base/models.py
```diff
@@ -398,12 +398,30 @@ class AbstractRadiusAccounting(models.Model):
...
-                                                     db_index=True,
                                                      # the default MySQL freeradius schema defines
                                                      # this as NOT NULL but defaulting to empty string
                                                      # but that wouldn't work on PostgreSQL
                                                      null=True,
                                                      blank=True)
+    framed_ipv6_address = models.GenericIPAddressField(verbose_name=_('framed IPv6 address'),
+                                                       db_column='framedipv6address',
+                                                       null=True,
+                                                       blank=True)
```
nemesisdesign: the ipv6 fields should have GenericIPAddressField.protocol set to ipv6:
https://docs.djangoproject.com/en/3.0/ref/models/fields/#genericipaddressfield

django_freeradius/base/models.py
```diff
@@ -398,12 +398,30 @@ class AbstractRadiusAccounting(models.Model):
...
+    framed_interface_id = models.CharField(verbose_name=_('framed interface ID'),
+                                           max_length=19,
+                                           db_column='framedinterfaceid',
+                                           null=True,
+                                           blank=True)
+    delegated_ipv6_prefix = models.CharField(verbose_name=_('delegated IPv6 prefix'),
+                                             max_length=44,
+                                             db_column='delegatedipv6prefix',
+                                             null=True,
+                                             blank=True)
```
nemesisdesign: can you please write a validator (eg: `ipv6_network_validator`) which ensures that the value, if not falsy, is an ipv6 subnet?

See django validators: 

- https://docs.djangoproject.com/en/3.0/ref/validators/

Use the ipaddress library (standard python 3 library):

- create an ip_network object: https://docs.python.org/3/library/ipaddress.html#ipaddress.ip_network
- if it fails with an exception, raise a validation error (eg: `raise ValidationError('Invalid ipv6 prefix: {}'.format(e))`)
- if the resulting object is not an instance of `IPv6Network` raise a validation error, eg: `raise ValidationError('{} is not an IPv6 prefix'.format(value))`

Write a test for these cases.

django_freeradius/base/models.py
```diff
@@ -398,12 +398,30 @@ class AbstractRadiusAccounting(models.Model):
...
                                                      blank=True)
+    framed_ipv6_address = models.GenericIPAddressField(verbose_name=_('framed IPv6 address'),
+                                                       db_column='framedipv6address',
+                                                       null=True,
+                                                       blank=True)
+    framed_ipv6_prefix = models.CharField(verbose_name=_('framed IPv6 prefix'),
+                                          max_length=44,
+                                          db_column='framedipv6prefix',
+                                          null=True,
+                                          blank=True)
```
nemesisdesign: also add the `ipv6_network_validator` here as well please?

django_freeradius/base/validators.py
```diff
@@ -0,0 +1,12 @@
...
+
+from django.core.exceptions import ValidationError
+
+
+def ipv6_network_validator(value):
+    try:
+        network = ip_network(value)
+    except Exception as e:
+        raise ValidationError('Invalid ipv6 prefix: {}'.format(e))
+    if not (isinstance(network, IPv6Network)):
```
nemesisdesign: I think you added an extra pair of parentheses which is not needed, should be:

```python
if not isinstance(network, IPv6Network):
```

PR #257, opened 2019-12-15
[Improvement] Migrate html2pdf to weasyprint

### LINKED ISSUE #243: [pdf] Migrate to weasyprint
Migrate the html2pdf library to wasyprint, which is way to superior to what we are using now.

nikitaermishin:
PR for improvement commit, which makes a migration from html2pdf to weasyprint. Changes:
 - Adds `weasyprint>=44` to requirements.txt instead of `xmtml2pdf`
 - Changes `STATIC_URL` in tests/settings.py to compatible with weasyprint
 - Creates `BATCH_PDF_TEMPLATE_BASE_URL` in django_freeradius/settings.py
 - Changes `generate_pdf` function in django_freeradius/utils.py, now it works using weasyprint API

Closes #243

### DIFF
--- a/django_freeradius/templates/django_freeradius/prefix_pdf.html
+++ b/django_freeradius/templates/django_freeradius/prefix_pdf.html
@@ -1,20 +1,30 @@
-<!DOCTYPE html>{% load static %}
+<!DOCTYPE html>{% load i18n %}
 <html>
 <head>
-    {% block style %}
-	<link href="{% static 'django-freeradius/css/freeradius.css' %}" rel="stylesheet" />
-    {% endblock %}
+    <title>{% trans 'WiFi username and passwords' %}</title>
+    <style type="text/css">
+        body {
+            font-size: 15px;
+            font-family: sans-serif
+        }
+        table { width: 100%; border-collapse: collapse }
+        table th { font-weight: bold }
+        table th, table td {
+            border: 1px solid #ccc;
+            padding: 10px;
+        }
+    </style>
 </head>
 <body>
 <table>
     <tr>
-        <th>Username</th>
-        <th>Password</th>
+        <th>{% trans 'Username' %}</th>
+        <th>{% trans 'Password' %}</th>
     </tr>
-    {% for u in users %}
+    {% for row in users %}
     <tr>
-	{% for x in u %}
-        <td>{{ x }}</td>
+	{% for cell in row %}
+        <td>{{ cell }}</td>
 	{% endfor %}
     </tr>
     {% endfor %}
--- a/django_freeradius/utils.py
+++ b/django_freeradius/utils.py
@@ -10,7 +10,7 @@ from django.core.validators import validate_email
 from django.template.loader import get_template
 from django.utils.crypto import get_random_string
 from django.utils.translation import ugettext_lazy as _
-from xhtml2pdf import pisa
+from weasyprint import HTML
 
 from django_freeradius.settings import BATCH_PDF_TEMPLATE
 
@@ -72,9 +72,9 @@ def prefix_generate_users(prefix, n, password_length):
 
 def generate_pdf(prefix, data):
     template = get_template(BATCH_PDF_TEMPLATE)
-    html = template.render(data)
+    html = HTML(string=template.render(data))
     f = open("{}/{}.pdf".format(settings.MEDIA_ROOT, prefix), "w+b")
-    pisa.CreatePDF(html.encode("utf-8"), dest=f, encoding="utf-8")
+    html.write_pdf(target=f)
     f.seek(0)
     return File(f)
 
--- a/requirements.txt
+++ b/requirements.txt
@@ -6,4 +6,4 @@ djangorestframework>=3.8.2,<3.10.0
 passlib>=1.7.1,<1.8.0
 django-filter>=2.1.0,<2.2.0
 djangorestframework-link-header-pagination>=0.1.1,<0.2.0
-xhtml2pdf>=0.2.3,<0.3.0
+weasyprint>=43,<51


django_freeradius/utils.py
```diff
@@ -72,9 +72,9 @@ def prefix_generate_users(prefix, n, password_length):
 
 def generate_pdf(prefix, data):
     template = get_template(BATCH_PDF_TEMPLATE)
-    html = template.render(data)
-    f = open("{}/{}.pdf".format(settings.MEDIA_ROOT, prefix), "w+b")
-    pisa.CreatePDF(html.encode("utf-8"), dest=f, encoding="utf-8")
+    html = HTML(string=template.render(data), base_url=BATCH_PDF_TEMPLATE_BASE_URL)
+    f = open("{}/{}.pdf".format(settings.MEDIA_ROOT, prefix), "wb+")
```
2stacks: Is this change necessary?

requirements.txt
```diff
@@ -6,4 +6,4 @@ djangorestframework>=3.8.2,<3.10.0
 passlib>=1.7.1,<1.8.0
 django-filter>=2.1.0,<2.2.0
 djangorestframework-link-header-pagination>=0.1.1,<0.2.0
-xhtml2pdf>=0.2.3,<0.3.0
+weasyprint>=44
```
2stacks: Probably should set an upper bound.

tests/settings.py
```diff
@@ -62,7 +62,8 @@
 USE_TZ = True
 USE_I18N = False
 USE_L10N = False
-STATIC_URL = '/static/'
+
+STATIC_URL = os.path.abspath('django_freeradius/static') + '/'
```
2stacks: Are these changes necessary?

django_freeradius/utils.py
nikitaermishin: No, will turn it back, thanks for the comment

requirements.txt
nikitaermishin: Fixed that

tests/settings.py
nikitaermishin: Yeah, they are necessary to achieve a correct path styles and js for the pdf

2stacks: So you're saying it doesn't work at all with `STATIC_URL = '/static/'`

django_freeradius/utils.py
```diff
@@ -72,9 +72,9 @@ def prefix_generate_users(prefix, n, password_length):
 
 def generate_pdf(prefix, data):
     template = get_template(BATCH_PDF_TEMPLATE)
-    html = template.render(data)
+    html = HTML(string=template.render(data), base_url=BATCH_PDF_TEMPLATE_BASE_URL)
```
nemesisdesign: why don't you use just `STATIC_URL` and avoid adding `BATCH_PDF_TEMPLATE_BASE_URL`?

I think adding the setting `BATCH_PDF_TEMPLATE_BASE_URL` is not necessary, if you think is necessary, please explain why.

tests/settings.py
nemesisdesign: this is wrong. `STATIC_URL` should be a relative URL, not an absolute path. Why have you changed this?
I believe you're doing something wrong if you need to do this.

nikitaermishin:
@nemesisdesign The problem is about providing `base_url` to `HTML` function. `base_url` should provide absolute path to the folder, which contains our `css` and `js` files. I suppose we could do it by some better way (including fix of `STATIC_URL` value mistake). So I can compose `BATCH_PDF_TEMPLATE_BASE_URL` of absolute way to `django_freeradius` folder and `STATIC_URL` variable. Will provide commit soon, so you will have opportunity to understand what I am talking about. :smiley:

nikitaermishin:
> I think this `weasyprint>=43,<=44` should be `weasyprint>=44,<45`

@2stacks It has no other versions between `44` and `45`, so isn't that the same as `weasyprint==44`? If so, could I use any of them?

2stacks:
> > I think this `weasyprint>=43,<=44` should be `weasyprint>=44,<45`
> 
> @2stacks It has no other versions between `44` and `45`, so isn't that the same as `weasyprint==44`? If so, could I use any of them?

Yes, they all achieve the same thing.  However, if you look at the current file, my RFC is purely based on consistency.  At some point 44 will be the supported lower bound and some future 4x will be the upper bound.

requirements.txt
```diff
@@ -6,4 +6,4 @@ djangorestframework>=3.8.2,<3.10.0
 passlib>=1.7.1,<1.8.0
 django-filter>=2.1.0,<2.2.0
 djangorestframework-link-header-pagination>=0.1.1,<0.2.0
-xhtml2pdf>=0.2.3,<0.3.0
+weasyprint>=43,<=44
```
nemesisdesign: The latest weasyprint release is 50, why are you putting 44 as the upper bound?
https://pypi.org/project/WeasyPrint/#history

nikitaermishin:
@nemesisdesign Ok, rebased my branch. Now working on the problem :smiley:

django_freeradius/utils.py
```diff
@@ -72,9 +73,10 @@ def prefix_generate_users(prefix, n, password_length):
 
 def generate_pdf(prefix, data):
     template = get_template(BATCH_PDF_TEMPLATE)
-    html = template.render(data)
+    html = HTML(string=template.render(data))
+    css = CSS(filename=os.path.abspath('django_freeradius/static/django-freeradius/css/freeradius.css'))
```
nikitaermishin: Is it a bad practice to import `os` and then use it only once in my code for this feature (line 77). Is it better to create a new variable in settings file (like `BATCH_PDF_TEMPLATE`) and then import it here?

nemesisdesign: this approach doesn't work, I tested it from the web interface of the development server and it raises an error.
it works when running tests with `runtests.py` because it happens that from the root, the path `django_freeradius/static/django-freeradius/css/freeradius.css` is correct, but from `tests/` it's not.

Moreover, if we hardcode the CSS value like that, we won't allow users to change their style.

I found a simpler approach. Get rid of the CSS entirely, we'll use an embedded stylesheet definition in the HTML it's not even really used. Then follow the instructions of my next comment.

django_freeradius/templates/django_freeradius/prefix_pdf.html
```diff
@@ -1,10 +1,6 @@
-<!DOCTYPE html>{% load static %}
+<!DOCTYPE html>
 <html>
-<head>
-    {% block style %}
-	<link href="{% static 'django-freeradius/css/freeradius.css' %}" rel="stylesheet" />
-    {% endblock %}
-</head>
+<head></head>
```
nemesisdesign: replace the contents of this file with the following:

```html
<!DOCTYPE html>{% load i18n %}
<html>
<head>
    <title>{% trans 'WiFi username and passwords' %}</title>
    <style type="text/css">
        body {
            font-size: 15px;
            font-family: sans-serif
        }
        table { width: 100%; border-collapse: collapse }
        table th { font-weight: bold }
        table th, table td {
            border: 1px solid #ccc;
            padding: 10px;
        }
    </style>
</head>
<body>
<table>
    <tr>
        <th>{% trans 'Username' %}</th>
        <th>{% trans 'Password' %}</th>
    </tr>
    {% for row in users %}
    <tr>
	{% for cell in row %}
        <td>{{ cell }}</td>
	{% endfor %}
    </tr>
    {% endfor %}
</table>
</body>
</html>
```

Users can supply a different template by using the `DJANGO_FREERADIUS_BATCH_PDF_TEMPLATE` setting.

nemesisdesign:
PS: please also rebase on the current master.

requirements.txt
```diff
@@ -6,4 +6,4 @@ djangorestframework>=3.8.2,<3.10.0
 passlib>=1.7.1,<1.8.0
 django-filter>=2.1.0,<2.2.0
 djangorestframework-link-header-pagination>=0.1.1,<0.2.0
-xhtml2pdf>=0.2.3,<0.3.0
+weasyprint>=43,<=50
```
nemesisdesign: here should be `weasyprint>=43,<51`

nikitaermishin: Ok, but why we should use `<51`? Is it because of subversions of `50` may be released (like `50.1`)?

nemesisdesign: yes, that's what may happen and is usually the case for most libraries

